PR #1, opened 2020-05-26
feat!: stub out rough structure for project

bcoe:
This gives us a little bit more of a starting point than an empty project:

* an example of building a service using express and postgres (_noting that there's a good chance we end up using firestore.).
* a relatively barebones React application, with testing, linting, a dev server, etc.

CC: @sofisl


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,57 @@
...
+  pull_request:
+name: ci
+jobs:
+  server-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [12]
+    services:
+      postgres:
```
JustinBeckwith, line 14: Postgres is an interesting choice :) Do you think Spanner, Bigtable, or Firestore are viable here?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,57 @@
...
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: flaky_dev_test
+        ports:
+        - 5432:5432
+        # needed because the postgres container does not provide a healthcheck
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    steps:
+      - uses: actions/checkout@v1
```
JustinBeckwith, line 25: Careful to use `checkout@v2`!

README.md
```diff
@@ -2,50 +2,23 @@
...
-
-```bash
-npm install flaky-service
-```
-
-
-
+## Developing
 
+* `docker-compose.yml`: run `docker-compose up`, to start the flaky.dev service
```
JustinBeckwith, line 18: Instead of compose, would it be possible to use scaffold and minikube?  These are a little more aligned to our strategy.

packages/frontend/__tests__/account-test.js
```diff
@@ -0,0 +1,70 @@
...
+
+  renderer.create(
+    <Router history={history}>
+      <PrivateRoute exact path="/account">
+        <Account/>
+      </PrivateRoute>
+    </Router>
+  );
+  setTimeout(() => {
+    expect(history.location.pathname).toBe('/');
```
JustinBeckwith, line 46: Where does `expect` come from exactly?  Any chance of using explicit imports?  Also, any chance of `assert` instead?

packages/frontend/src/css/index.css
```diff
@@ -0,0 +1,28 @@
...
+
+.home .header {
+  background-color: black;
+  min-height: 55px;
+  margin-bottom: 10px;
+  font-family: arial;
+  font-size: 22pt;
+  color: white;
+  padding: 8px 4px 4px 4px;
+}
```
JustinBeckwith, line 28: Gimme that sweet line ending

packages/server/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "start": "node index.js",
+    "lint": "xo",
+    "fix": "xo --fix",
+    "migrate:create": "db-migrate create --config=./database.js",
+    "migrate:up": "db-migrate up --config=./database.js",
+    "migrate:down": "NODE_ENV=development db-migrate down --config=./database.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/bcoe/flaky.dev.git"
```
JustinBeckwith, line 18: Need to update this :)

packages/server/package.json
```diff
@@ -0,0 +1,56 @@
...
+  },
+  "keywords": [
+    "flkay",
+    "dev"
+  ],
+  "author": "Google LLC",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/bcoe/flaky.dev/issues"
```
JustinBeckwith, line 27: And this

packages/server/package.json
```diff
@@ -0,0 +1,56 @@
...
+  },
+  "homepage": "https://github.com/bcoe/flaky.dev#readme",
```
JustinBeckwith, line 29: and this

packages/server/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "express-session": "^1.17.0",
+    "pg": "^7.18.2",
+    "pino": "^5.17.0",
+    "simple-oauth2": "^3.3.0",
+    "uuid": "^7.0.2",
+    "xml2js": "^0.4.23"
+  },
+  "devDependencies": {
+    "c8": "^7.1.0",
+    "chai": "^4.2.0",
```
JustinBeckwith, line 45: For consistency, could we not `chai`?

synth.py
```diff
@@ -29,4 +29,3 @@
   '.kokoro/continuous/node10/system-test.cfg',
```
JustinBeckwith, line 29: Do we want a synthfile here?

.github/workflows/ci.yaml
bcoe: I think Firestore would be good to explore using.

README.md
bcoe: I threw this in last night so that someone could at least consider running the project locally... I expect we'll throw out postgres for firestore in the first week or two of the project...

Basically as soon as I can get a Pantheon account setup. tldr; I don't know if we should worry too much about Docker related choices right now.

packages/frontend/__tests__/account-test.js
bcoe: expect is injected in by Jest, I believe. It sounds like we're leaning towards using angular any ways, having talked to the team, which means we'll likely not be using Jest.

bcoe: we will make sure our test imports are explicit when we make the switch.

synth.py
bcoe: Let's hold off on `synth.py` for now, I'd rather use actions.

PR #7, opened 2020-06-10
feat: integrated gts

gis727:
I installed, set up and ran GTS linter.


bcoe:
@gis727 great work, thanks for getting us consistent with Google's TSC conventions 😄

PR #8, opened 2020-06-11
feat: add a build to Firestore and compute percent passing

StuartRucker:
Add a build to Firestore and compute the passing percentage for the build and each test case.

### DIFF
--- a/packages/api/lib/testrun.js
+++ b/packages/api/lib/testrun.js
@@ -1,3 +1,17 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 class TestCaseRun {
   constructor (okMessage, number, name) {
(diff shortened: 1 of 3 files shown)


packages/api/lib/testrun.js
```diff
@@ -0,0 +1,15 @@
...
+  constructor (okMessage, number, name) {
+    this.successful = (okMessage === 'ok');
+    this.number = number;
+    this.name = name;
+    this.encodedName = encodeURIComponent(this.name);
+    this.failureMessage = 'TODO ERROR MESSAGE, (e.g. stackoverflow error line 13)';
+  }
+
+  display () {
+    return this.number + ', ' + this.name + ', ' + this.time + ', ' + (this.successful ? '1' : '0');
```
bcoe, line 25: for ease of shipping this through a UI, we'll probably want to eventually make this a data-structure, rather than a string, so something like:

```
{
  successes: 3,
  failures: 5,
  timeMs: 4994
}
```

packages/api/test/add-build-test.js
```diff
@@ -0,0 +1,185 @@
...
+
+const TESTING_COLLECTION_BASE = 'repositories-testsuite-';
+
+// The three builds that will be added
+const buildInfo = [
+  {
+    repoId: encodeURIComponent('nodejs/node'),
+    organization: 'nodejs',
+    buildId: '11111',
+    url: 'https://github.com/nodejs/WRONG', // URL starts off wrong
```
bcoe, line 30: good call mocking out the builds, this is exactly the approach I would have suggested 👍

packages/api/lib/testrun.js
StuartRucker: This class represents a singular run of a test case, but what you said still applies

cedpeters:
Looks great!

PR #9, opened 2020-06-12
feat: create the home and the search components

gis727:


gis727:
The landing page is now the HomeComponent and the search component can lookup some mock repositories.
Screenshot

packages/frontend/src/app/home/home.component.css
```diff
@@ -20,4 +20,23 @@ limitations under the License.
...
+#search-bar {
+	margin-top: 10px;
+	display: flex;
+	width: 100%;
+	align-content: center;
+	justify-content: center;
+}
+
+app-search {
+    width: 100%; /* Take the entire space on mobile devices */
```
bcoe, line 40: not to be too nit-picky, but I think we're using different spacing than the CSS above.

packages/frontend/src/app/home/home.component.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  it("should have as title 'flaky-dashboard'", () => {
+    expect(component.title).toEqual('flaky-dashboard');
+  });
+
+  it('should render title', () => {
+    fixture.detectChanges();
+    const compiled = fixture.nativeElement;
+    expect(compiled.querySelector('span').textContent).toContain(
+      'flaky-dashboard is running !'
```
bcoe, line 55: Assertions like this will over time need to keep being updated, not a blocker for this PR, but it might be worth looking at [snapshot testing](https://izifortune.com/snapshot-testing-angular-applications/) eventually.

packages/frontend/src/app/search/search.component.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    }).compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(SearchComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
```
bcoe, line 39: If possible, I would love to see a test that demonstrates that interacting with the search component cause a change to the page.

Will it update the home component with results, or redirect to a new page with results?

packages/frontend/src/app/services/com/com.service.ts
```diff
@@ -0,0 +1,30 @@
...
+import {HttpClient} from '@angular/common/http';
+import {apiLinks} from './api';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class COMService {
+  constructor(private http: HttpClient) {}
+
+  public fetchRepositories(): Promise<string[]> {
```
bcoe, line 25: Lovely, @madelyndubuk this provides a demonstration of how we can talk form the frontend to the server-side 😄

madelyndubuk: Oh yes I see, this is great!

packages/frontend/src/app/search/search.component.spec.ts
gis727: Will proceed ! The interactions are on the way !
On this page, it will redirect to the repository/organisation page.
But on the organisation page, it will simply update the shown list.

PR #10, opened 2020-06-12
feat: added Firebase/Cloud Run deployment files

madelyndubuk:


packages/frontend/.firebase/hosting.ZGlzdA.cache
```diff
@@ -0,0 +1,23 @@
+404.html,1591747591833,b7bab6b83fa074653ff28c8d2a64135d3434575f70a12ab3d3ba8080461b9537
+flaky-dashboard/favicon.ico,1591746399596,5ba04cf43d3c8c026bd5edd43046db96acb28b5686d99ea7d7e148feabfffc8f
+flaky-dashboard/index.html,1591746399596,9c65f8872da56c1dba66e92931fd6ee05b5dfcdbbc3f63b0e922b99d70ea4717
+flaky-dashboard/main-es2015.js,1591746364694,67f5510d8c7cf7199bc13a3bac7af11dd6e5c5c0caccf5d327208c7840c00d2e
+flaky-dashboard/main-es2015.js.map,1591746364694,ca51fc7528135e48b38fc8ee9f25c93c89cc20bfb6b487281edd4c5762c63aa9
+flaky-dashboard/main-es5.js,1591746364694,dc2e89dbe2cc314a51ff11054ffd6eaf1decb1680283df2504a4e55d314f1d1a
+flaky-dashboard/main-es5.js.map,1591746364694,0d0662076088644344a2eeeb71bc8166d72570a40baf0ca868b05a0ba0f53df3
+flaky-dashboard/polyfills-es2015.js,1591746364694,e0ca47e773d9520e5ed5e267c3e5f63c5299d0fa3080fcd8e4f3acf287d2db97
```
bcoe, line 8: we should probably add the `.firebase` folder to `.gitignore` in the top level, I think you can just add the line:

```
.firebase
```

and then run `git rm --force packages/frontend/.firebase`.

packages/api/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# Copy application dependency manifests to the container image.
+# A wildcard is used to ensure both package.json AND package-lock.json are copied.
+# Copying this separately prevents re-running npm install on every code change.
+COPY package*.json ./
+
+# Install production dependencies.
+# Use the official lightweight Node.js 12 image.
+# https://hub.docker.com/_/node
+FROM node:12-slim
```
bcoe, line 16: we should only need to pull in the `node:12-slim` instance once at the top, I think this second `FROM` will be a noop.

packages/api/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# Create and change to the app directory.
+WORKDIR /usr/src/appRUN npm install --only=production
+
+
+# Copy local code to the container image.
+COPY . ./
```
bcoe, line 23: We shouldn't need two `COPY` commands either, looks almost like two Dockerfiles might have been added on top of each other 😄 

I think we can pretty much almost use [this exact file](https://github.com/knative/docs/blob/ca0a1b32da4df27a043ebdbd0bd032d1723bbadb/docs/serving/samples/hello-world/helloworld-nodejs/Dockerfile) which it looks like you're close to.

madelyndubuk: Oh yes you are right, so the second COPY command should be removed? The other file you attached has a COPY ../ command and performs COPY package*.json ./ separately

bcoe: Having multiple copy commands is fine, but I think that `COPY .`  would actually copy over `package.json`, maybe?

If things stop working after that change I'm wrong 😆

packages/frontend/.firebaserc
```diff
@@ -0,0 +1,5 @@
+{
+  "projects": {
+    "default": "oss-automation"
```
bcoe, line 3: we'll want to make this a variable eventually, so that it's easy to deploy to other projects.

PR #12, opened 2020-06-15
feat: adding /repos endpoint

cedpeters:
Add /repos and tests for the added methods. @bcoe and @StuartRucker


packages/api/test/server-test.js
```diff
@@ -34,10 +34,18 @@ describe('flaky express server', () => {
...
+    var sol = ['firstRepo', 'fourthRepo', 'secondRepo', 'thirdRepo'];
+    const respJSON = await resp.json()
+    assert.deepEqual(respJSON.repoNames, sol);
+  });
+  
   it('it returns a single repository, when you call GET on /repository/:id');
   it('it creates a repository, when you call POST on /repository');
   after(() => {
-    server.close();
+    // server.close();
```
bcoe, line 49: I think tests are never finishing because we're missing this close after all the tests.

packages/api/test/repository-test.js
```diff
@@ -15,12 +15,20 @@
 const { describe, it, after, before } = require('mocha');
 const Repository = require('../src/repository');
 const assert = require('assert');
+const fetch = require('node-fetch');
```
bcoe, line 18: looks like this `fetch` dependency isn't used.

In practice, I'd try to keep use of `node-fetch` to the `server-test.js`, and have tests that run against models not perform HTTP requests.

packages/api/server.js
```diff
@@ -14,11 +14,43 @@
...
 const express = require('express');
+const Repository = require('./src/repository');
 const app = express();
 const bodyParser = require('body-parser');
 
 app.use(bodyParser.json());
 
+app.get('/repos', async (req, res) => {
+  // TODO: Make it return more information about the repos, beyond just their names
+  var repository = new Repository(null);
```
bcoe, line 25: nit: but I suggest using `const` rather than `var`, if variables aren't going to be reassigned.

packages/api/server.js
```diff
@@ -14,11 +14,43 @@
...
+  var result = await repository.getCollection('dummy-repositories');
+
+  let repoNames = [];
+
+  for (var index = 0; index < result.length; index++) {
```
bcoe, line 30: nit: `let` rather than `var`...

Here's a blog post [on the topic](https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75) basically, you want to opt for `let` and `const` rather than `var` these days:

* `let`: if you need to reassign.
* `const`: if you do not.

packages/api/server.js
```diff
@@ -14,11 +14,43 @@
...
+    const id = result[index].repositoryid;
+    // console.log(index + ": " + id);
+    repoNames.push(id);
+  }
+
+  repoNames = ['firstRepo', 'fourthRepo', 'secondRepo', 'thirdRepo'];
```
bcoe, line 36: I like that you're starting with fake data, this is a very pragmatic approach.

packages/api/src/repository.js
```diff
@@ -37,6 +37,28 @@ class Repository {
...
+      .then(snapshot => {
+        if (snapshot.empty) {
+          console.log('No matching documents.');
+          return;
+        }
+
+        snapshot.forEach(doc => {
+          var entry = doc.data();
+          result.push(entry);
+        // console.log(doc.id, '=>',entry);
```
bcoe, line 52: I usually drop comments that aren't needed any longer.

bcoe:
see: https://github.com/GoogleCloudPlatform/flaky-service/issues/14

packages/api/server.js
```diff
@@ -14,11 +14,43 @@
...
+  const result = await repository.getCollection('dummy-repositories');
+
+  const repoNames = [];
+
+  for (let index = 0; index < result.length; index++) {
+    const id = result[index].repositoryid;
+    repoNames.push(id);
+  }
+
+  // TODO: Comment out the below if the data in the database changes so that the test starts failing
```
bcoe, line 35: I think we can drop this comment now.

StuartRucker:
LGTM

PR #13, opened 2020-06-16
feat: receive post of a build

StuartRucker:
An API endpoint to receive a post request with information on a build and put it in FireStore


packages/api/server.js
```diff
@@ -27,9 +30,11 @@ app.post('/', (req, res) => {
     message: req.body.message ? req.body.message : 'hello world'
   });
 });
+var postBuildHandler = new PostBuildHandler(app, client);
+postBuildHandler.listen();
 
 const port = process.env.PORT ? Number(process.env.PORT) : 3000;
-const server = app.listen(port, () => console.log(`Example app listening at http://localhost:${port}`));
+const server = app.listen(port, 'localhost', () => console.log(`Example app listening at http://localhost:${port}`));
```
bcoe, line 37: I would avoid specifying `localhost` here, this will actually cause trouble when we deploy to production.

Is letting it bind to the default port causing issues on your computer?

packages/api/server.js
```diff
@@ -27,9 +30,11 @@ app.post('/', (req, res) => {
     message: req.body.message ? req.body.message : 'hello world'
   });
 });
+var postBuildHandler = new PostBuildHandler(app, client);
```
bcoe, line 33: I would use `const` rather than `var` (if you're never going to reassign the value).

I like this approach of pulling the route binding into a class, I recommend coordinating with @cedpeters and using this approach for other routes.

packages/api/test/post-build-test.js
```diff
@@ -0,0 +1,115 @@
...
+const fetch = require('node-fetch');
+
+describe('flaky express server', () => {
+  let server;
+  let client;
+  before(async () => {
+    const serverModule = require('../server');
+    server = serverModule.server;
+    client = new Firestore();
+    global.headCollection = 'repositories-testsuite-' + Math.random().toString(36).substr(2, 9);
```
bcoe, line 30: I'd pull in the [uuid](https://www.npmjs.com/package/uuid) dependency, and use this to generate a UUID, rather than `Math.random()`.

If you like math, [here's a good read on why](https://medium.com/@betable/tifu-by-using-math-random-f1c308c4fd9d) 😆

packages/api/test/post-build-test.js
```diff
@@ -0,0 +1,115 @@
...
+
+const assert = require('assert');
+const fetch = require('node-fetch');
+
+describe('flaky express server', () => {
+  let server;
+  let client;
+  before(async () => {
+    const serverModule = require('../server');
+    server = serverModule.server;
```
cedpeters: This is unnecessary. Just set:
const server = require('../server');

PR #15, opened 2020-06-17
feat!: /api/build endpoint, github action

StuartRucker:
Breaking Change: Added /api prefix to postbuild routes, and adjusted urls.

Added an additional endpoint + tests to send a raw TAP file as input instead of a parsed JSON object. In the packages/action folder, index.js sends raw TAP file and index-deprecated.js sends the parsed TAP file


.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 .vscode
 node_modules
+!action/node_modules
```
bcoe, line 4: to avoid having to check in the `node_modules` folder, I think it might be worth looking at adding a build step that uses [ncc](https://help.github.com/en/actions/creating-actions/creating-a-javascript-action).

packages/action/action.yaml
```diff
@@ -0,0 +1,20 @@
+name: 'Post TAP Files'
```
bcoe, line 1: Let's call this thing `flaky.dev` 😄

packages/action/action.yaml
```diff
@@ -0,0 +1,20 @@
...
+description: 'Greet someone and record the time'
+inputs:
+  github:
+    description: 'The Github Context JSON object'
```
bcoe, line 5: I think most of this information we need is populated in environment variables, see:

https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

So, the user shouldn't need to provide this information.

packages/action/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "action",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
```
bcoe, line 9: Let's add a linter to this library, like we have for the API. I suggest we use the same one:

https://github.com/GoogleCloudPlatform/flaky-service/blob/master/packages/api/package.json#L9

This can then just be our `test`.

packages/action/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "license": "ISC",
```
bcoe, line 10: License should be `Apache-2.0`.

packages/action/package.json
bcoe, line 9: we can use the author `Google LLC`.

packages/api/package.json
```diff
@@ -28,6 +28,8 @@
     "@google-cloud/firestore": "^3.8.4",
     "body-parser": "^1.19.0",
     "express": "^4.17.1",
+    "stream": "0.0.2",
```
bcoe, line 31: Readable Stream is built in to Node.js core now, so you shouldn't need this module:

https://nodejs.org/api/stream.html#stream_class_stream_readable

packages/api/src/post-build.js
```diff
@@ -66,6 +92,19 @@ class PostBuildHandler {
         res.status(400).send({ error: err.message });
       }
     });
+
+    // route for parsing test input server side
+    this.app.post('/postbuildraw', async (req, res, next) => {
```
bcoe, line 97: I'd be tempted to call this just `POST /build`, if we end up having something other than a raw upload in the future, we can use the HTTP header `Content-Type` to specify the different types.

packages/action/action.yaml
StuartRucker: The environment variables are used, but I think they have to be used in the .github/workflows file, and this action.yaml metadata file just passes these environment variables into the javascript

PR #18, opened 2020-06-19
feat!: better handling of environment data + api prefix

StuartRucker:
BREAKING-CHANGE: added /api prefix to remaining / and /repos routes

GET api/build/:repoid?parameters
GET api/buildenv/:repoid 

Added these two paths to learn about the possible parameters relating to a build (os, ref, tag, matrix) and then return builds sorted chronologically relating to that path.

Added the tests to same file as the add-build-test.js file to use the same database information, and changed the filename to test/addbuild-getbuild.js.

Manually added the 2^4-1=15 compound indices for sorting permutations of these four fields

Not done yet: Handle searching based on git sha.


packages/api/test/server-test.js
```diff
@@ -19,7 +19,7 @@ const fetch = require('node-fetch');
 
 describe('flaky express server', () => {
   it('it responds to a GET on the / path', async () => {
-    const resp = await fetch('http://localhost:3000', {
+    const resp = await fetch('http://localhost:3000/api', {
```
cedpeters, line 22: nit: I think we're supposed to replace 'localhost' with '0.0.0.0'

cedpeters, line 22: nit: we should change 'localhost' to '0.0.0.0' at some point.

packages/api/src/add-build.js
```diff
@@ -60,11 +60,14 @@ function buildPassingPercent (successes, failures) {
...
 
-  // if repo does not exist, add information
-  await dbRepo.update({
-    organization: buildInfo.organization,
-    url: buildInfo.url
-  }, { merge: true });
+  // if repo does not exist, add information including environment variables and possibilities
+  const repoUpdate = { };
+  for (const prop in buildInfo.environment) {
+    repoUpdate['environments.' + prop] = Firestore.FieldValue.arrayUnion(buildInfo.environment[prop]);
```
bcoe, line 66: are there any illegal characters we should be concerned about, what happens for instance if the property is `foo.bar`, will we accidentally create an extra level of nesting in FireStore?

I wonder if `Firestore` provides a method for escaping strings, or has documentation on doing so.

packages/api/src/add-build.js
```diff
@@ -90,11 +93,13 @@ async function addBuild (testCases, buildInfo, client, collectionName = 'reposit
     // 2: Query an appropriate amount of test runs, for now 100 most recent
     var snapshot = await dbRepo.collection('tests').doc(testCase.encodedName).collection('runs').orderBy('timestamp').limit(100).get();
 
-    // 3: Compute the passing percent and write that back
-    await dbRepo.collection('tests').doc(testCase.encodedName).update({
-      environments: Firestore.FieldValue.arrayUnion(buildInfo.environment),
-      percentpassing: testCasePassingPercent(snapshot)
-    }, { merge: true });
+    // 3: Compute the passing percent and write that back as well as updated environment variables
```
bcoe, line 96: I appreciate the comments outlining your algorithms, this is exactly the sort of thing comments are great for.

packages/api/src/get-build.js
```diff
@@ -0,0 +1,59 @@
...
+
+class GetBuildHandler {
+  constructor (app, client) {
+    this.app = app;
+    this.client = client;
+  }
+
+  listen () {
+    // return the different environments for a build - all parameters and possible values for next query
+    this.app.get('/api/buildenv/:repoid', async (req, res, next) => {
```
bcoe, line 25: I think I said this in another PR, I like the approach you're taking of encapsulating routes inside classes 😄 

@cedpeters we're probably going to want to do the same for the auth work you're doing.

packages/api/test/addbuild-getbuild.js
```diff
@@ -0,0 +1,267 @@
...
+const { describe, it, after, before } = require('mocha');
+const { Firestore } = require('@google-cloud/firestore');
+const assert = require('assert');
+const { v4: uuidv4 } = require('uuid');
+
+const TestCaseRun = require('../lib/testrun');
+const addBuild = require('../src/add-build');
+const fetch = require('node-fetch');
+
+const TESTING_COLLECTION_BASE = 'repositories-testsuite-';
```
bcoe, line 24: this approach is going to start to cause problems when the project gets more active, we have a few branches happening at the same time, and potentially a few test runs happening at the same time.

An approach I've seen work well is to use a UUID in the collection name, so something like:

```
const TESTING_COLLECTION_BASE = `repositories-testsuite-${uuidv4}`;
```

You would also need to:

1. delete the `TESTING_COLLECTION_BASE` after the test run.
2. have an additional cleanup step that looks for collections with the name `repositories-testsuite*` older than N hours, and cleans these up.

☝️ mind creating an issue to do this work, I'm fine with it not being part of the suite yet.

packages/api/test/addbuild-getbuild.js
```diff
@@ -0,0 +1,267 @@
...
+
+  describe('getBuild', async () => {
+    it('Can get parameters for the build', async () => {
+      const resp = await fetch('http://localhost:3000/api/buildenv/nodejs%2Fnode');
+      const respJSON = await resp.json();
+      const sol = { organization: 'nodejs', environments: { matrix: [{ 'node-version': '12.0' }], os: ['linux-apple', 'linux-banana'], tag: ['abc', 'xyz'], ref: ['master'] }, url: 'https://github.com/nodejs/node' };
+      assert.deepStrictEqual(respJSON, sol);
+    });
+
+    it('Can get limit and sort by date', async () => {
```
bcoe, line 223: It's exciting to see these test cases get fleshed out, great work.

packages/api/test/addbuild-getbuild.js
StuartRucker: This was currently being done a few lines down: 
`global.headCollection = TESTING_COLLECTION_BASE + uuidv4();`

packages/api/src/add-build.js
StuartRucker: Fixed with firebaseEncode function in lib/firebase-encode.js and test/lib-test.js

PR #19, opened 2020-06-20
feat: search-communication

gis727:
- Repositories can be fetched from the server (tested on a local server with CORS enabled).
- 200 ms debouncing in the search bar.
- An interpreter for the search filters is present now and will be used during the redirection to the MainPage.

Here is a not-so-descriptive image of a few repositories fetched from the server:
The repositories aren't filtered yet on the server.
Screenshot 2020-06-19 at 8 31 36 PM


packages/api/server.js
```diff
@@ -20,7 +20,9 @@ const bodyParser = require('body-parser');
 const PostBuildHandler = require('./src/post-build.js');
 const { Firestore } = require('@google-cloud/firestore');
 const client = new Firestore();
+const cors = require('cors');
 
+app.use(cors());
```
bcoe, line 25: @cedpeters @StuartRucker this here just needs to be added to allow the browser to talk to the backend API.

packages/frontend/src/app/search/interpretation/interpretation.service.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+      expect(filter).not.toBeUndefined();
+
+      // the filters should have been lowercased
+      const splittedInput = input.toLowerCase().split(service.filterSeparator);
+      expect(filter.name).toEqual(splittedInput[0]);
+      expect(filter.value).toEqual(splittedInput[1]);
+    });
+  });
+
+  it('should not parse wrong filters', () => {
```
bcoe, line 125: thank you for continuing to write good tests, I can't underscore how much this will be important to the future health of this project 😄

PR #21, opened 2020-06-22
feat: license component

madelyndubuk:
- added license component to bottom right of screen
- license button in side navigation bar

Untitled document

Button in side navigation bar:

Untitled document (1)

Untitled document (2)


bcoe:
@madelyndubuk bother you to add a screen shot for what this looks like now?

PR #25, opened 2020-06-23
feat: land-main-component

gis727:
Landed the component that renders repositories list. This design is subject to changes based on folks feedback.
loc

Integrated that component and the search bar inside the MainComponent.
glob

Also:
- Added routing with a few filters support from the landing page to the repositories list page.
- Added search support in the repositories list page (mock repositories are used at this time because the server isn't ready yet).
- The repositories list is now functional (the user can move through the different pages).


bcoe:
this is looking really good 👏 , I dig it.

bcoe:
@gis727 a little bit of a UX nit, I might make the whole background of a repo with flaky tests slightly yellow, and the whole background of a repo with failing tests slightly red, I think that would jump out more than the word `flaky`.

gis727:
Nice catch ! Will do !

PR #26, opened 2020-06-23
feat!: switch to passing ID in query string for the time being

StuartRucker:
Breaking change: Switched the repoid parameter from a URL string parameter to a query parameter

api/build/:repoid -> api/build?repoid=..


gis727:
> This seems like a good approach top unblock us 👍 two things:
> 
> 1. could we open up a tracking issue to move back to the `/:repoid` form at some point in the future, since this is more true to form for REST design.
> 2. could we coordinate with @gis727 and also update the query on the frontend in this PR, it will need to change its form too I think (to use a GET parameter rather than URL).

We aren't using this route yet in the frontend. But the communication protocol will be updated.

StuartRucker:
* Opened an issue to do this using repoid in domain
* this path not used in frontend yet

PR #30, opened 2020-06-23
feat!: api/repo, api/org, and  api/repo/{{org}}/{{name}} endpoints

StuartRucker:
/api/repo returns all repos and has optional query parameters of startswith (for either repo name, or org/reponame)
/api/org/{{orgname}} returns all repos for this org
/api/repo/{{org}}/{{repo}} replaces /api/build and api/buildenv, and returns {metadata: <<info about repo>>, builds: <a list of builds>}  (BREAKING CHANGE)

Now supports case insensitive querying

Change also includes storing the repo name, org name, and repoid as fields in firestore both normally and in lower case for the case insensitive query

Changed Some error codes from 400->500


packages/api/src/get-repo-orgs.js
```diff
@@ -0,0 +1,97 @@
...
+          const snapshotOrg = await this.client.collection(global.headCollection)
+            .where('repoId', '>=', req.query.startswith)
+            .where('repoId', '<', this.alphabeticallyIncrement(req.query.startswith))
+            .limit(limit).get();
+          snapshotOrg.docs.forEach(doc => results.push(doc.data()));
+
+          res.send(this.removeDuplicates(results));
+        }
+      } catch (error) {
+        res.status(400).send({
```
bcoe, line 70: for an unknown error, we should probably opt for a `500`.

packages/api/src/get-repo-orgs.js
```diff
@@ -0,0 +1,97 @@
...
+            .limit(limit).get();
+          snapshotName.docs.forEach(doc => results.push(doc.data()));
+
+          const snapshotOrg = await this.client.collection(global.headCollection)
+            .where('repoId', '>=', req.query.startswith)
+            .where('repoId', '<', this.alphabeticallyIncrement(req.query.startswith))
+            .limit(limit).get();
+          snapshotOrg.docs.forEach(doc => results.push(doc.data()));
+
+          res.send(this.removeDuplicates(results));
```
bcoe, line 67: I think we might be able to simplify this if we assume that we know the organization, so are just querying for the repos under that organization. I think we could assume that on the front end, someone has indicated the org they're interested in before hitting this route.

packages/api/src/get-repo-orgs.js
```diff
@@ -0,0 +1,97 @@
...
+        newResults.push(result);
+      }
+    });
+    return newResults;
+  }
+
+  // TODO order by flakyness
+  listen () {
+    // return the different environments for a build - all parameters and possible values for next query
+    this.app.get('/api/reposearch', async (req, res, next) => {
```
bcoe, line 45: In the spirit of REST, I would probably opt for `/api/repo` for this:

```
GET /api/repo.  -- list of all repos.
GET /api/repo/name -- some specific repo.
POST /api/repo -- create a repo.
PUT /api/repo -- update a repo.
```

We don't have the `POST` and `PUT` endpoints, because this data gets populated by your `/build` endpoint.

packages/api/package.json
```diff
@@ -5,7 +5,7 @@
   "description": "the api for the flaky.dev service",
   "main": "server.js",
   "scripts": {
-    "test": "c8 mocha --timeout=25000 test/*.js",
+    "test": "c8 mocha --timeout=25000 test/get-repoorgs-test.js test/setup.js",
```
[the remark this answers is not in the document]
StuartRucker: Oh thanks! This is helpful.

packages/api/src/get-repo-orgs.js
StuartRucker: I was imagining initially it would be a repositorycentric view, so you would want to search for repositories from the landing page by typing strings like "Googlecloudplatform/fl" or "flaky-se" or "Goog"

packages/api/src/get-repo-orgs.js
StuartRucker: Fixed, with second route being api/repo/:org/:name

gis727:
Not blocking:
In **/api/org/{{orgname}}** it would be nice to have the org metadata too (will be used for the page presenting the heat map).

packages/api/src/get-build.js
```diff
@@ -58,9 +44,12 @@ class GetBuildHandler {
...
         const resp = [];
         snapshot.forEach(doc => resp.push(doc.data()));
-        res.send(resp);
+
+        var metadata = await this.client.collection(global.headCollection).doc(firebaseEncode(repoid)).get();
+
+        res.send({ metadata: metadata.data(), builds: resp });
       } catch (err) {
-        res.status(400).send({ error: err.message, stack: err.stack, name: err.name });
+        res.status(500).send({ error: err.message, stack: err.stack, name: err.name });
```
bcoe, line 52: Not a blocker right now,  but eventually we probably want to obfuscate the stack trace, unless we're in development mode.

You tend to cleanup errors before serializing them to the frontend, because they can contain sensitive information about how your system work.

packages/api/src/get-build.js
```diff
@@ -14,36 +14,22 @@
...
-        const results = await this.client.collection(global.headCollection).doc(encodeURIComponent(req.query.repoid)).get();
-        res.send(results.data());
-      } catch (err) {
-        res.status(400).send({ error: err });
-      }
-    });
-
     // return the recent builds based on parameters
-    this.app.get('/api/build', async (req, res, next) => {
+    this.app.get('/api/repo/:orgname/:reponame', async (req, res, next) => {
```
bcoe, line 27: I like where we landed on this URL pattern 👍

PR #35, opened 2020-06-25
feat: flaky analytics

StuartRucker:
Computes whether builds, repositories, or tests are flaky.

A **test** is flaky if both...
- there has been a failure within the last 5 runs
- there have been any two failures separated by between 1 and 10 successes
Note: These sequences are based entirely on the timestamp of the build

A **build** is Flaky if a flaky test failed on this build

A **repository** is flaky if it contains at least one flaky test. This means that a flaky repository could contain a flaky build as long as this is not the most recent build.

In the case of builds being added non chronologically:
- The repository's numfails and numtests does not change and still reflects the most recent build
- Each test case's flaky status is updated, still based on the most recent chronological builds. (This means no change if the build is far back or not close chronologically to another failure)
- The repository's flaky count updates based on the test cases
- The build's flaky count is based only on chronologically prior builds.

**API Change:** /api/repo & api/repo/:org/:repo name now include numfails, numtests, and flaky (integer representing # flaky tests) values


packages/api/test/analytics-test.js
```diff
@@ -0,0 +1,185 @@
...
+      'tests/{testcase}',
+      'builds/{buildid}'
+    ];
+    const buildIds = ['1', '2', '3', '4'];
+    const testCaseIds = [];
+    for (let j = 0; j < testCases.length; j++) {
+      testCaseIds.push(j.toString());
+    }
+    // Delete all possible documents, okay to delete document that doesnt exist
+    deletePaths.forEach(async (deletePath) => {
```
bcoe, line 174: Mixing `forEach` and `async` can be a bit risky, because promises eventual values. I think this will end up creating N promises immediately, not delete test data in sequence.

If you do want to delete several items in parallel, it can be more idiomatic to go:

```
await Promise.all(testCaseIds.map(async () => {
     const deletePathUse = deletePath.replace('{testcase}', encodeURIComponent(testCase)).replace('{buildid}', buildId);
      await client.collection(global.headCollection).doc(buildInfo.repoId + '/' + deletePathUse).delete();
});
```

If you want to delete in sequence, you might do better to unravel this into a for statement.

PR #41, opened 2020-06-30
feat: Authentication

cedpeters:
This includes session tracking, and a /githubauth endpoint which will redirect the user to log in with Github and do all of the back and forths for the login process.


packages/api/server.js
```diff
@@ -56,6 +71,65 @@ app.get('/api/repos', async (req, res) => {
...
+
+  request.post(
+    {
+      headers: { 'content-type': 'application/json', Accept: 'application/json' },
+      url: 'https://github.com/login/oauth/access_token',
+      body: JSON.stringify({
+        code: req.param('code'),
+        client_id: process.env.CLIENT_ID,
+        client_secret: process.env.CLIENT_SECRET,
+        state: 'MUSTCHANGE' // TODO: make this something better
```
gis727, line 93: Not blocking:
Maybe save _MUSTCHANGE_ it in a const ? (since its at line 75 and 80)

cedpeters: Yes, this will likely ultimately be a uuid value.

packages/api/server.js
```diff
@@ -56,6 +71,65 @@ app.get('/api/repos', async (req, res) => {
...
+          const permitted = await repository.mayAccess('github', JSON.parse(body).login);
+          await repository.storeSessionPermission(req.sessionID, permitted);
+          res.status(200).redirect('https://flaky-dev-staging.web.app');
+        }
+      );
+    }
+  );
+});
+
+app.get('/sessionauthcheck', async (req, res) => {
```
gis727, line 123: @cedpeters , i think we want **/api/sessionauthcheck** instead.

packages/api/server.js
```diff
@@ -56,6 +71,65 @@ app.get('/api/repos', async (req, res) => {
     .end();
 });
 
+app.get('/githubauth', (req, res) => {
```
gis727, line 74: **/api/githubauth** ?

packages/api/server.js
```diff
@@ -56,6 +71,65 @@ app.get('/api/repos', async (req, res) => {
...
+        },
+
+        await async function (error, response, body) {
+          if (error || JSON.parse(body).message === 'Bad credentials') {
+            res.status(401).redirect('https://flaky-dev-staging.web.app');
+            return;
+          }
+          const repository = new Repository();
+          const permitted = await repository.mayAccess('github', JSON.parse(body).login);
+          await repository.storeSessionPermission(req.sessionID, permitted);
```
bcoe, line 115: would the `permitted` variable be saved automatically, if you were to set:

```
req.session.permitted
```

This is the API I've seen before.

packages/api/server.js
```diff
@@ -56,6 +71,65 @@ app.get('/api/repos', async (req, res) => {
...
+          res.status(200).redirect('https://flaky-dev-staging.web.app');
```
bcoe, line 116: we should probably make this an environment variable too:

```
const redirect = process.env.FRONTEND_URL || 'http://127.0.0.1:4222
```

packages/api/server.js
```diff
@@ -56,6 +71,65 @@ app.get('/api/repos', async (req, res) => {
...
+        }
+      );
+    }
+  );
+});
+
+app.get('/sessionauthcheck', async (req, res) => {
+  const repository = new Repository();
+  const result = await repository.sessionPermissions(req.sessionID);
```
bcoe, line 125: I would be tempted to call this `/session`, and then you could return the `username` currently logged in, which could be used to display this contextual information to the frontend.

packages/api/test/server-test.js
```diff
@@ -18,6 +18,48 @@ const assert = require('assert');
...
+    it('redirects to the flaky-dev home page when finished (regardless of how auth goes)', async () => {
+      const resp = await fetch('http://0.0.0.0:3000/api/callback?state=MUSTCHANGE', {
+        headers: { redirect: 'manual' }
+      });
+      assert.strictEqual(resp.url, 'https://flaky-dev-staging.web.app/');
+    });
+  });
+
+  describe('/sessionauthcheck', async () => {
+    // it('returns a JSON object indicating whether the user is authenticated', async () => {
```
bcoe, line 56: I would remove these tests if they're not something we want to assert against yet.

packages/api/src/repository.js
```diff
@@ -61,8 +60,51 @@ class Repository {
     return result;
   }
 
-  async delete (identifier) {
-    const document = this.client.doc(`${REPOSITORY_COLLECTION}/${identifier}`);
+  async mayAccess (platform, login) {
+    let solution;
+    const collection = this.client.collection('permitted-users/' + platform + '/users').where('login', '==', login);
+
+    await collection.get().then(function (querySnapshot) {
```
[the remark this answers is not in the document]
cedpeters: Ah okay thanks! I didn't know about this, but makes sense. I just pushed a fix.

packages/api/server.js
```diff
@@ -33,6 +36,18 @@ global.headCollection = process.env.HEAD_COLLECTION || 'testing-buildsget';
...
 app.use(cors());
 app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({ extended: true })); // support encoded bodies
+app.use(
+  session({
+    store: new FirestoreStore({
+      dataset: new Firestore(),
+      kind: 'express-sessions'
+    }),
+    secret: 'my-secret',
```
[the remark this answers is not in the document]
cedpeters: Should all sessions have the same secret, which is stored in Secret Manager? Or should I generate a new uuid for each one and store it?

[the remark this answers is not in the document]
bcoe: We should have one secret that's used to encrypted the session, this will be different than the `github-client-secret`.

[`SESSION_SECRET`](https://www.npmjs.com/package/express-session#secret) is a secret we control, which is used to sign the session delivered to the frontend, such that folks can't easily mess with the data we store in the session.

packages/api/test/repository-test.js
```diff
@@ -12,51 +12,78 @@
...
+    });
+
+    it('returns not permitted if there is no expiration date', async () => {
+      await repo.createDoc('express-sessions/test-session', {});
+      const permission = await repo.sessionPermissions('test-session');
+      assert.deepStrictEqual(permission, { permitted: false, expiration: null, login: null });
+      await repo.deleteDoc('express-sessions/test-session');
+    });
+
+    it('returns not permitted if the expiration date is in the past');
```
bcoe, line 87: we should create a tracking issue to fill in this test eventually 👍

packages/api/test/server-test.js
```diff
@@ -12,12 +12,56 @@
...
+process.env.FRONTEND_URL = 'https://flaky-dashboard.web.app/home';
 const { describe, it } = require('mocha');
 
 const assert = require('assert');
 const fetch = require('node-fetch');
 
 describe('flaky express server', () => {
+  const frontendUrl = 'https://flaky-dashboard.web.app/home';
+
+  it('mock environment variables work', () => {
```
bcoe, line 25: nit: I usually try to make these read like a sentence:

```
it('should have mocked environment variables')
```

packages/api/package.json
```diff
@@ -25,12 +25,18 @@
...
     "body-parser": "^1.19.0",
     "cors": "^2.8.5",
     "express": "^4.17.1",
+    "express-session": "^1.17.1",
+    "mocked-env": "^1.3.2",
+    "moment": "^2.27.0",
+    "querystring": "^0.2.0",
     "tap-parser": "^10.0.1",
-    "uuid": "^8.1.0"
+    "url": "^0.11.0",
```
bcoe, line 38: you shouldn't need a `url` dependency, as a URL parser is built into Node.js.

packages/api/package.json
```diff
@@ -25,12 +25,18 @@
   },
   "homepage": "https://github.com/GoogleCloudPlatform/flaky-service#readme",
   "dependencies": {
+    "@google-cloud/connect-firestore": "^1.0.6",
     "@google-cloud/firestore": "^3.8.4",
     "body-parser": "^1.19.0",
     "cors": "^2.8.5",
     "express": "^4.17.1",
+    "express-session": "^1.17.1",
+    "mocked-env": "^1.3.2",
```
bcoe, line 34: it looks like you didn't end up using this, so we can just remove this:

```
npm uninstall mocked-env
```

packages/api/package.json
```diff
@@ -25,12 +25,18 @@
...
+    "moment": "^2.27.0",
+    "querystring": "^0.2.0",
```
bcoe, line 36: `querystring` is built into Node.js, like `url`:

https://nodejs.org/api/querystring.html

There should be no need for the dependency.

PR #46, opened 2020-06-30
feat: config panel

### LINKED ISSUE #42: Add a configuration dialog
- [ ] should be accessible in a right-sidenav under a wheel on every page
- [ ] should have an option to:
  - [x] add (link) a github account
  - [x] see the api token
  - [ ] renew the api token

gis727:
fixes #42 

Added a wheel button to the upper-right corner of the viewport
BIG

Added configs in the sidenav. The conditional access to the configs are provided by #69.
The git accounts have been removed.
FOCUS


gis727:
This is a potential starting point for @cedpeters
The sidenav can be as wide as the viewport if necessary.

bcoe:
@gis727 for the MVP of config I've been discussing with @cedpeters, I don't think we'll need the ability to link GitHub accounts, rather there will be a login button.

I think this is creating a good foundation 😄 could you try staging it with the branch @madelyndubuk has configured?

PR #47, opened 2020-07-01
feat: formal routing

### LINKED ISSUE #43: Add the repository builds view
- [x] should show every build with pagination
- [x] should have filters

### LINKED ISSUE #51: Set formal routing
- [x] **/:org** should land on the search page searching for all repositories in **org**
- [x] **/search** should work without any parameter
- [x] **/:org/:repo** should land on the repo page searching for all builds in **org/repo**
- [x] Use **/search;repo=** instead of **/search;query=**
- [ ] All filters should be sanitized

gis727:
fixes #51 
- The interpretation service (src/app/services/interpretation) handles the new routing format.
- The pagination logic is extracted into an abstract component (src/app/paginated-list/paginated-list.component.ts).
- Added a utility service (src/app/services/utils.service.spec.ts).
- Updated the main.component and the repo-list.component with those new features.


packages/frontend/src/app/services/utils.service.ts
```diff
@@ -0,0 +1,36 @@
...
+
+@Injectable({
+  providedIn: 'root',
+})
+export class UtilsService {
+  constructor(public sanitizer: DomSanitizer) {}
+
+  parseUrl(url: string): SafeUrl {
+    // TODO: should be done on the server ideally
+    url = url?.replace('git://', 'https://');
```
bcoe, line 27: I think the logic for turning a `git://` address into its repository address is a bit more nuanced:

https://www.npmjs.com/package/github-url-from-git

gis727: @StuartRucker do you think we can do this in the api ? That way the client receives links only.

packages/frontend/src/app/main/repo-list/repo-list.component.ts
```diff
@@ -12,44 +12,22 @@
...
-  _repositories: Repository[] = [];
   @Input() set repositories(value: Repository[]) {
-    this._repositories = value;
+    this._elements = value;
     this.updatePage();
     this.paginator?.firstPage();
   }
-
-  renderedRepositories: Repository[] = [];
-  pageIndex = 0;
```
StuartRucker, line 35: This unmerged PR adds support for server side pagination. on the /api/repo endpoint there is a limit, startsaftername, and endsbeforename parameter, so if you want to go to the next page pass in startsaftername as an input with the last name on the current page. 

https://github.com/GoogleCloudPlatform/flaky-service/pull/54

packages/frontend/src/app/services/interpretation/interpreters/route-interpreter.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export interface FoundParams {
+  queries: Map<string, string>;
+  filters: Filter[];
+}
+
+export const expectedParams: Map<string, ExpectedParams> = new Map([
+  ['main', {queries: ['org', 'repo'], filters: []}],
+  ['repo', {queries: ['org', 'repo'], filters: ['tag', 'matrix', 'ref', 'os']}],
+  ['build', {queries: ['org', 'repo', 'build'], filters: ['order by']}],
```
StuartRucker, line 66: Let's discuss the possible orderings. Right now, the only thing supported is ordering by timestamp which it does by default.

packages/frontend/src/app/main/repo-list/repo-list.component.ts
gis727: Got it 👍
Knowing that, my strategy was to fetch a max number of data (ie: 1000) and when/if the user reaches the last page (or a page close to the last one), the client will request more data based on the name of the last repo.
**In other words, startsaftername and endsbeforename will be perfect !** 👌

packages/frontend/src/app/services/interpretation/interpreters/route-interpreter.ts
gis727: Yes !
Maybe also order by status (failing 1st, flaky 2nd and the rest 3rd (each subgroup ordered alphabetically) ).

PR #54, opened 2020-07-07
feat!: build info + paginate repos on org specific query

StuartRucker:
Breaking Change: /repo now requires org parameter

Changed the data model so that a build now not only stores all test cases, but also stores 
- passing percent
- whether its flaky
for each test.

The repo query was modified to require an organization, and to support pagination.

Note: In the future there may need to be some feature designed to disable the "next" and "previous" button if there are no more results. However, there currently could be workarounds such as increasing the "limit" parameter by one, and not displaying the last result but using it to determine whether there is more.


packages/api/src/add-build.js
```diff
@@ -125,15 +126,21 @@ async function addBuild (testCases, buildInfo, client, collectionName = 'reposit
     // update information for flakybuild and flakyrepo;
     flakyBuild = (buildtestCaseAnalytics.computeIsFlaky() && !testCase.successful) ? 1 + flakyBuild : flakyBuild;
     flakyRepo = (updateObj.flaky) ? 1 + flakyRepo : flakyRepo;
+    alltests.push({
+      name: testCase.name,
+      passed: testCase.successful,
+      flaky: updateObj.flaky,
```
bcoe, line 132: If we change our algorithm for flakiness over time, is it a problem that we'd have historic tests that were flagged as flaky using the old approach?

packages/api/test/get-repoorgs-test.js
```diff
@@ -95,31 +90,33 @@ describe('Getting Repos and Orgs', () => {
...
-    for (let t = 0; t < querySuffixes.length; t++) {
-      const resp = await fetch('http://localhost:3000/api/repo?startswith=' + querySuffixes[t]);
-      const respText = await resp.text();
-      assert.strictEqual(JSON.parse(respText).length, 4);
-    }
-  });
-  it('should return all with no parameter', async () => {
-    const resp = await fetch('http://localhost:3000/api/repo');
+  // deprecated feature:
+  // it('should be able to search by org/beginning of name', async () => {
```
bcoe, line 99: I tend to be pretty cavalier about deleting code, rather than commenting it out (_the git history can always be referenced if we want to look at an old feature._).

packages/api/src/add-build.js
StuartRucker: Yes, when looking at a particular historical build, it's tests will be marked according to their status at the time of the build being posted. This means that even when you look at a historical build, it may be listed as flaky even if it has since been fixed. Note that metrics for a repository and for an individual test case will be updated.

The potential solutions are
1. Keeping this and when viewing a build see (Flaky as of this build, percent passing up to this build)
2. Not display this information for a particular build since it could be potentially confusing.
3. When viewing a particular build, see the most up to date flaky status and percent passing status. This overall seems better but has potential to be confusing. 

To accomplish (3.), it will involve no longer storing a list of all tests for a particular build. Then, when getting a particular build, iterating through the list of all test cases and querying .doc(<test name>).collection('runs').doc(<buildId>)

packages/api/test/paginate-query-test.js
```diff
@@ -0,0 +1,110 @@
...
+
+const assert = require('assert');
+const fetch = require('node-fetch');
+
+const data = [
+  // name, org
+  ['aa', 'alpha'],
+  ['ab', 'alpha'],
+  ['ac', 'alpha'],
+  ['b', 'alpha'], // test numbers
```
tbpg, line 29: What does this comment mean?

packages/api/src/add-build.js
bcoe: we discussed this in chat:

I'm comfortable with `1.`, since builds represent a historical snapshot of the state of the tests at the time of the build. Whereas, if I'm looking at tests in aggregate, it should represent the current state of tests.

packages/api/test/paginate-query-test.js
StuartRucker: Left over comment. Good catch.

PR #57, opened 2020-07-07
feat: filter

gis727:
Added a filter component

Screenshot 2020-07-08 at 6 50 13 PM


gis727:
👍. There is a discussion going on about it at the end of #47

PR #59, opened 2020-07-08
feat: build page

### LINKED ISSUE #48: Add the test view
- [x] should show every test for a specific build with pagination.
- [x] for each test:
  - [x] should show id, name.
  - [x] should allow to see the context (build | action) on GitHub.
- [x] for each failing test, should show the failure message.

gis727:
**Note** : Files in packages/frontend/src/app/filters are in #57 (not to be reviewed here)

Fixes #48 

Added the build page (shows the tests for a specific build)
build

 - Failing test:
failing_test

  - Flaky test:
flaky_test

  - Passing test:
successful_test


packages/frontend/src/app/app-routing.module.ts
```diff
@@ -17,13 +17,15 @@ import {Routes, RouterModule} from '@angular/router';
...
 import {MainComponent} from './main/main.component';
+import {BuildComponent} from './build/build.component';
 
 const routes: Routes = [
   {path: '', redirectTo: '/home', pathMatch: 'full'},
   {path: 'home', component: HomeComponent},
   {path: 'license', component: LicenseComponent},
   {path: 'search', component: MainComponent},
   {path: ':org', component: MainComponent},
+  {path: ':org/:repo/:build', component: BuildComponent},
```
bcoe, line 28: I think @chingor13 brought this up today, we should probably have a prefix like:

```
/org/:org/:repo/:build
```

Otherwise, what if an org named `home` is using our product?

packages/frontend/src/app/build/build.component.ts
```diff
@@ -0,0 +1,126 @@
...
+    {
+      name: 'should update the rendered pages on input change',
+      flaky: true,
+      failing: true,
+      percentpassing: 98,
+      environment: {os: 'windows', ref: 'dev'},
+    },
+    {
+      name:
+        'should not return to the first page when the paginator is not ready',
```
bcoe, line 54: I love the way you're mocking API responses 😄 this is very pragmatic testing.

packages/frontend/src/app/app-routing.module.ts
gis727: Ok, switching to that. I should mention that in the marketing PR, we are dropping "/home" to use "/" though.

packages/frontend/src/app/build/tests-list/test-details/test-details.component.html
```diff
@@ -0,0 +1,73 @@
...
+<mat-divider></mat-divider>
+
+<!-- FAILURE DETAILS -->
+<div *ngIf="test.failing" class=".outer-field">
+    <section>
+        <!-- HEADER -->
+        <header id="error-header">Error message</header>
+
+        <!-- MESSAGE -->
+        <textarea readonly id="error-msg-textarea" rows="5" >Error with ...&#013;&#010;Unable to ...
```
bcoe, line 42: How does this interaction work now, is it a pop up like in the screen shot?

I'm happy with this as a starting point, but to @JustinBeckwith's point, I can imagine we make an effort to avoid pop ups eventually 😄

gis727: Yes, it is.
We will use an expansion eventually instead.

PR #60, opened 2020-07-08
feat: 404 page

gis727:
Added the 404 page.
That page is shown when a requested link doesn't match any known pattern.


packages/frontend/src/app/not-found/not-found.component.html
```diff
@@ -0,0 +1,27 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<div id="main-container">
+    <h1>404</h1>
+
+    <h3>Sorry, but the thing you are looking for isn't here. You can:</h3>
```
madelyndubuk, line 20: Just a suggestion for wording in place of "the thing you are looking for isn't here": "The page you are looking for does not exist" or the "The page you are looking for is not found" ?

packages/frontend/src/app/not-found/not-found.component.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    ).nativeElement;
+
+    const locationBackSpy = spyOn(location, 'back');
+
+    headBackAnchor.click();
+
+    expect(locationBackSpy).toHaveBeenCalledTimes(1);
+  });
+
+  it('should head to the home page when user clicks on the corresponding anchor', fakeAsync(() => {
```
bcoe, line 68: nit: I'd probably use the term, "should redirect".

PR #63, opened 2020-07-09
feat: added folders for staging and production deployment instructions

madelyndubuk:
- Added a nested directory structure to distinguish between staging and production environment
- Now we should be able to push to the testing branch, view the staging site, and then make a PR from that testing branch


packages/frontend/deployment/firebase.json
```diff
@@ -1,7 +1,7 @@
 {
   "hosting": {
     "target": "prod",
-    "public": "dist/flaky-dashboard",
+    "public": "staging/dist/flaky-dashboard",
```
bcoe, line 4: I would be tempted to use the paths:

* `frontend/deployment/staging/*`, and.
* `frontend/deployment/production/*.

Rather than having the production config live one folder above the staging config.

madelyndubuk: I agree that is a better approach.  The problem I found with that is there is only one `dist/` folder which the firebase project needs access to and it must be at the same level or at a level below in order for the `firebase.json` file to read it.  I can play around with some other options to make it better but when I tried having two staging and production folders, there wasn't a place for the `dist` folder where both files could read

madelyndubuk:
Got it! I will look into it now.

PR #64, opened 2020-07-09
fix: add 'org/' support in redirections

gis727:
Added 'org/' support in redirections on repositories and builds page.


packages/frontend/src/app/main/repo-list/repo-list.component.html
```diff
@@ -19,7 +19,7 @@
         class="repo"
         [class.flaky-repo]="!repo.numfails && repo.flaky"
         [class.failing-repo]="repo.numfails"
-        (click)="onElementClick(repo.repoId, $event)">
+        (click)="onElementClick(location.path() + '/' + repo.name, $event)">
```
chingor13, line 22: This seems a bit fragile. We should be using angular's build-in routing for buliding urls. Generating these paths ourselves makes this component change the url depending upon where it's being used.

If we choose to refactor the routing, we will have to remember all the places we manually generate the urls.

PR #65, opened 2020-07-09
Feat: Validate POST requests to /api/build

StuartRucker:
Use the github app installation token from the Github action described [here](https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/#authenticating-as-an-installation) to validate all posts to /api/build.

Tested using [Nock](https://www.npmjs.com/package/nock)


packages/api/lib/validate-github.js
```diff
@@ -0,0 +1,51 @@
...
+    return false;
+  }
+
+  const respJSON = await resp.json();
+
+  if (!respJSON.repositories.length) {
+    return false;
+  }
+
+  for (let i = 0; i < respJSON.repositories.length; i++) {
```
bcoe, line 40: you can do:

```js
for (const repo of repoJSON.repositories) {
  repo.full_name
}
```

packages/api/test/res/sample-validate-resp.json
```diff
@@ -0,0 +1,88 @@
...
+         "private":false,
+         "owner":[
+            "Object"
+         ],
+         "html_url":"https://github.com/StuartRucker/mocha-chai-project",
+         "description":"This is a sample Mocha Chai testing project.",
+         "fork":true,
+         "url":"https://api.github.com/repos/StuartRucker/mocha-chai-project",
+         "forks_url":"https://api.github.com/repos/StuartRucker/mocha-chai-project/forks",
+         "keys_url":"https://api.github.com/repos/StuartRucker/mocha-chai-project/keys{/key_id}",
```
bcoe, line 19: I like the fixture based approach, we do this a lot in our automation repo.

packages/api/src/post-build.js
```diff
@@ -105,24 +106,18 @@ class PostBuildHandler {
...
     this.app.post('/api/build', async (req, res, next) => {
       try {
-        var buildInfo = this.parseBuildInfo(req.body.metadata);
-        var parsedRaw = await this.parseRawOutput(req.body.data, req.body.type);
-        var testCases = this.parseTestCases(parsedRaw);
+        const buildInfo = this.parseBuildInfo(req.body.metadata);
+        const parsedRaw = await this.parseRawOutput(req.body.data, req.body.type);
+        const testCases = this.parseTestCases(parsedRaw);
+
+        const isValid = await validateGithub(req.body.metadata.github.token, req.body.metadata.github.repository);
```
cedpeters, line 116: Once you've used the token to check validity, how long do you keep it? Are we storing it in the database as part of the metadata?

packages/api/test/post-build-test.js
```diff
@@ -32,10 +37,29 @@ describe('Posting Builds', () => {
       projectId: process.env.FLAKY_DB_PROJECT || 'flaky-dev-development'
     });
     global.headCollection = 'repositories-testsuite-' + uuidv4();
+
+    nock('https://api.github.com', {
```
cedpeters, line 41: I use node fetch in my tests. I suppose consistency between different test files probably isn't very important, though.

packages/api/src/post-build.js
StuartRucker: It is never stored. An api request is made using it and that is all.

packages/api/test/post-build-test.js
StuartRucker: Nock is a mock http endpoint module so when you make a request to the github api to validate a token you dont need a real token. Instead you can use nock to make calling a node fetch to that url returns whatever you want.

cedpeters: Oh okay I thought it was just an equivalent to node-fetch but that sounds awesome!

PR #66, opened 2020-07-09
fix: Firestore connect fix

cedpeters:
This is a small PR which fixes the issue where multiple connections are being made to Firestore, and some were trying to connect to OSS-Automation instead of the flaky database.


packages/api/src/add-build.js
```diff
@@ -18,8 +18,8 @@
 * and builds
 * returns analytics object and adds such data to database
 */
-const { Firestore } = require('@google-cloud/firestore');
 const { buildPassingPercent, TestCaseAnalytics } = require('../lib/analytics.js');
+const Firestore = require('@google-cloud/firestore');
```
StuartRucker, line 22: Are these things doing the same thing? I was copying [this](https://github.com/googleapis/nodejs-firestore), but that certainly looks cleaner.

packages/api/src/repository.js
```diff
@@ -12,28 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-const { Firestore } = require('@google-cloud/firestore');
 const moment = require('moment');
+const client = require('./firestore.js');
```
StuartRucker, line 16: if what you are trying to avoid is running the New Firestore() code multiple times, than this is not fixing that. It is my understanding that everytime you run require it will generate a new client object. The design pattern would be generating client once and passing that in in functions.

packages/api/src/add-build.js
cedpeters: As far as I can tell they are the same, the brackets are useful for when more than one variable is set by the requirement.

packages/api/src/repository.js
cedpeters: Good catch, I hadn't fully tested that! I just did by adding print statements in firestore.js, and in repository.js and server.js right after each of them required firestore.js, and while print statements from repository and server both appeared on the console, the print statement in firestore only printed once. This seems to show that the code is only run the first time firestore.js is called, and after that it simply exports the calculation it already did.

PR #67, opened 2020-07-10
feat: Updated populate script to reflect current data model

StuartRucker:
Updated the script that populates the database to include all of the information added to the data model since the first week.

Adds many repositories to the database with one test and random names to test out the searching, and Adds tons of tests and builds to the nodejs/node repository to test out subpages for a particular repository.

No tests, but the results of the script can be viewed [in Firestore](https://pantheon.corp.google.com/firestore/data/testing-buildsget) and accessed via the [Production API](https://flaky-dashboard.web.app/api)


StuartRucker:
> This is really useful for populating testing data. One thought, given the way this project has evolved.
> 
> Would it make sense for us to switch to loading a bunch of TAP test runs, from a variety of test runs? I'd be comfortable checking in a `data/` directory that has a few dozen `.tap` output files.
> 
> This is not a blocker for this PR.

Currently it looks in the testing_logs folder for TAP files. The only substantial examples I have are from the nodejs project so I did not upload them, but that is what I ran it with locally.

PR #70, opened 2020-07-10
feat: added the login guard

gis727:
**/login** now redirects to the api login endpoint


packages/frontend/e2e/src/app.e2e-spec.ts
```diff
@@ -30,10 +30,14 @@ describe('workspace-project App', () => {
   afterEach(async () => {
     // Assert that there are no errors emitted from the browser
     const logs = await browser.manage().logs().get(logging.Type.BROWSER);
-    expect(logs).not.toContain(
-      jasmine.objectContaining({
-        level: logging.Level.SEVERE,
-      } as logging.Entry)
-    );
+    logs.forEach(log => {
+      // TODO: Launch the server before e2e testing ?
```
gis727, line 34: We can set up a testing server for e2e testing or just mock the backend (making it not **_really_** e2e): https://jasonwatmore.com/post/2020/03/10/angular-8-fake-backend-example-for-backendless-development

bcoe: @StuartRucker has been using [nock](https://www.npmjs.com/package/nock) for low level mocking of network calls.

I like this approach because you intercept right at the network call, which gives me a lot of confidence that everything is working well -- I'd prefer this to an E2E test honestly.

bcoe: I wonder how hard it would be to perform E2E testing on the UI, but with `nock` to mock our session dance?

PR #72, opened 2020-07-12
feat: added a route provider

gis727:
Added a route provider to centralize route names definitions


packages/frontend/src/app/main/main.component.ts
```diff
@@ -39,7 +40,7 @@ export class MainComponent implements OnInit {
     this.route.params.subscribe(params => {
       const foundParams = this.interpreter.parseRouteParam(
         params,
-        expectedParams.get('main')
+        expectedParams.get(RouteProvider.routes.main.name)
```
gis727, line 43: The idea is to hold those strings in the provider.

packages/frontend/src/app/routing/app-routing.module.ts
```diff
@@ -22,14 +22,14 @@ import {BuildComponent} from '../build/build.component';
 import {NotFoundComponent} from '../not-found/not-found.component';
 import {apiLinks} from '../services/com/api';
 import {LoginGuard} from './route-guards/LoginGuard';
+import {RouteProvider} from './route-provider/RouteProvider';
 
 const routes: Routes = [
   {path: '', component: HomeComponent},
   {path: 'license', component: LicenseComponent},
-  {path: 'search', component: MainComponent},
```
gis727, line 29: As the organisation name will have to be known by the user, the search route isn't relevant.

packages/frontend/src/app/main/main.component.ts
bcoe: Is the goal here that we could display two types of  entities in the same main component, e.g., repositories in the repo view,  and tests  in the test view?

packages/frontend/src/app/repository/repository.component.ts
```diff
@@ -87,7 +88,7 @@ export class RepositoryComponent implements OnInit {
     this.route.params.subscribe(params => {
       const foundParams = this.interpreter.parseRouteParam(
         params,
-        expectedParams.get('repo')
```
bcoe, line 90: I'm a supporter of moving away from magic variables.

packages/frontend/src/app/routing/route-provider/RouteProvider.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+      args.push(arg);
+      argsString += '/' + arg;
+    }
+    return {args, argsString};
+  };
+
+  it('should return a route link corresponding to the provided parameters', () => {
+    Object.keys(RouteProvider.routes).forEach(route => {
+      const routeLinkProvider: Function = RouteProvider.routes[route].link;
+      const {args, argsString} = getLinkArgs(routeLinkProvider.length);
```
bcoe, line 33: I'm finding this test a little hard to follow, it might be worth hardcoding the link that's expected, rather than using the `getLinkArgs` which has quite a bit of logic for a test.

I believe we're trying to approve that a link will be created with the appropriate org and `orgName` and `repoName` populated? I think just showing the static link might be fine `/org/foo-org/bar-repo`.

packages/frontend/src/app/main/main.component.ts
gis727: The goal is actually to "untie" each component from a particular route. So the routes are all defined in the RouteProvider and each component can be linked to any of them. This is actually the first step in introducing a refactor requested by @chingor13 in #64.
The idea of reusing the same component for every page is coming in another PR after Martin's input.

PR #73, opened 2020-07-13
feat: added breadcrumbs

gis727:
Added the breadcrumbs.
Fixes #56

86672418-a15ddd80-bfc4-11ea-806e-f77786f7ca8a


packages/frontend/src/app/app-component/app.component.html
```diff
@@ -33,6 +33,10 @@
     </mat-sidenav>
 
     <mat-sidenav-content>
+      
+      <!-- BREADCRUMBS -->
```
bcoe, line 37: nit: comment might not be needed, since the code below is self documenting ... the problem with comments like this is that we might eventually  decide to use an alternate component, and leave the comment ... future generations will have:

```html
 <!-- BREADCRUMBS -->
 <app-not-a-bread-crumb />
```

packages/frontend/src/app/breadcrumbs/breadcrumbs.component.css
```diff
@@ -0,0 +1,48 @@
...
+    margin-top: 5px;
+}
+
+#home-icon {
+    font-size: 1.2em;
+    vertical-align: middle;
+    display: inline-flex;
+    align-items: center;
+    padding-bottom: 3px;
+    width: 18px;
```
bcoe, line 30: To @chingor13's drive by comment the other day, I would be curious if there are more off the shelf styles we can use.

Mind making this a topic of conversation when you have a 1:1 with Martin?

packages/frontend/src/app/routing/app-routing.module.ts
```diff
@@ -28,8 +28,23 @@ const routes: Routes = [
   {path: '', component: HomeComponent},
   {path: 'license', component: LicenseComponent},
   {path: 'org/:org/:repo/:build', component: BuildComponent},
-  {path: RouteProvider.routes.repo.path, component: RepositoryComponent},
```
bcoe, line 31: The breadcrumb path is reflected in the URL as well, i.e.,

We'd be at the URL: `/org/GoogleCloudPlatform/flaky-dev`, and have the breadcrumb `GoogleCloudPlatform -> flaky-dev`.

packages/frontend/src/app/breadcrumbs/breadcrumbs.component.css
gis727: Sure will !

PR #74, opened 2020-07-13
feat: static search bar

gis727:
- The search bar is now fixed (as the user scrolls) and appears on pages where an organisation name can be inferred from the url (main page, and repo page).
- Redirection on search validation is now handled by the search component itself (instead of the parent component).

Screenshot 2020-07-13 at 7 25 29 PM


packages/frontend/src/app/search/search.component.ts
```diff
@@ -71,15 +99,25 @@ export class SearchComponent implements OnInit {
   }
 
   onEnterKeyUp(option: string): void {
-    this.searchOptionSelected.emit(this.interpreter.parseSearchInput(option));
+    this.launchSearch(this.interpreter.parseSearchInput(option));
```
gis727, line 102: The search bar handles the redirection itself instead of moving the responsibility to the parent component.

PR #75, opened 2020-07-13
feat: added license component to main page and repo page

madelyndubuk:
- added license component to main org page and repo page


packages/frontend/src/app/repository/repository.component.ts
```diff
@@ -103,6 +106,15 @@ export class RepositoryComponent implements OnInit {
     });
   }
 
+  openLicenseDialog(): void {
+    const dialogConfig = new MatDialogConfig();
```
bcoe, line 110: since the height and width are consistent, would it make sense to define the width/height in CSS rather than as a variable?

madelyndubuk: I can look into doing it with CSS! I just have to figure out how that works since the dialog box container itself is not in any html file.  I will see if this can be done.

gis727: I think that those arguments for the dialog are required by the dialog module (https://material.angular.io/components/dialog/api#MatDialogConfig).
But would sure be nice to have them in the css.

madelyndubuk: That is what I found as well.  However I removed them and the default dimensions looked good to me when I added the `mat-dialog-content` container.  Here is a screenshot of the new size, let me know what you think @gis727 @bcoe 
Screenshot 2020-07-13 at 4 25 09 PM

PR #78, opened 2020-07-14
chore: added the provider in the repo list template

gis727:
Addresses @chingor13 comment in #64.
The only other template where this could be useful is _src/app/repository/repository.component.html_, but that wont be necessary at this time because of the new design we are going to land.


bcoe:
@gis727 looks like I caused a merge conflict for you, mind rebasing when you have a sec?

PR #79, opened 2020-07-14
chore: added tests list in the repository page

gis727:
Screenshot 2020-07-14   at 12 34 44 PM


packages/frontend/src/app/repository/repository.component.ts
```diff
@@ -32,43 +32,49 @@ import {RouteProvider} from '../routing/route-provider/RouteProvider';
   styleUrls: ['./repository.component.css'],
 })
 export class RepositoryComponent implements OnInit {
-  // mock builds until the data is fully formatted
-  mockBuilds = [
+  mockTests: Test[] = [
```
StuartRucker, line 35: Would these fields be enough? Currently I am not storing the environment for each build, but I could add it. I will add a new api endpoint to get the tests for the most recent build.

flaky: boolean
Name:String
Passed: boolean
percent passing: float (for last 30 builds)
status: "OK" or "error message

packages/frontend/src/app/repository/repository.component.html
```diff
@@ -50,10 +50,8 @@
     </div>
 </div>
 
-
-<!-- REPOSITORIES LIST -->
-<div class="build-list-container">
-    <app-build-list [builds]="mockBuilds" [repoName]="repoName" [orgName]="orgName"></app-build-list>
```
StuartRucker, line 56: Let's discuss details of which builds it will return, and whether we will leave scrolling / pagination until after we have MVP. BTW the api endpoint `/repo/:orgname/:reponame` should include all the information used for this page. Looking at response.builds[0].tests would return the limited test information for the part below

packages/frontend/src/app/repository/repository.component.ts
gis727: LGTM !

packages/frontend/src/app/repository/repository.component.html
gis727: Yes. Having the tests in _builds[0].tests_ makes sense. LGTM.
Yes, I think that we can leave the pagination for after the MVP. @bcoe ? No pagination here meaning that all tests will be fetched once (the pagination in the frontend will still be there).

PR #80, opened 2020-07-14
feat: Better TAP Parsing

StuartRucker:
Improved the parsing of TAP files so it can detect the failure message, as well as handle nested TAP. The following are now supported:

```
1..2
    1..2
    ok 1 - true is ok
    ok 2 - doag is also okay
ok 1 - nesting
ok 2 - second
```

and 

```
1..3
ok 1 Testing Box duplicatetest
ok 2 Testing Box should assert volume of the box to be 6000
not ok 3 Testing Box should throw an error
  AssertionError: expected 1 to equal 2
      at Context.<anonymous> (test/Box.test.js:20:12)
      at processImmediate (internal/timers.js:456:21)
```


packages/api/src/post-build.js
```diff
@@ -105,20 +135,50 @@ class PostBuildHandler {
...
+      });
+
+      // write your multiline variable to the child process
+      process.stdin.write(datastring);
+      process.stdin.end();
+    });
+    return output.replace(/\d+..\d+\n/, '\n'); // hacky solution to issue where 1..0 shows up at beginning
+  }
+
+  static removeDuplicateTestCases (testCases) {
```
tbpg, line 154: Does this take into account class names? For example, two different directories may have tests with the same name.

packages/api/src/post-build.js
```diff
@@ -105,20 +135,50 @@ class PostBuildHandler {
...
+    const testCasesNew = [];
```
tbpg, line 155: I usually name this something like `result`. That way, it's clear you're building this list up and plan to return it.

packages/api/src/post-build.js
StuartRucker: Most TAP generators would prefix the test names to be `class/test name`, but other than that there is no way to know since this endpoint accepts just one TAP file from all classes aggregated without class information

packages/api/lib/testrun.js
```diff
@@ -20,11 +20,11 @@ class TestCaseRun {
     this.number = number;
     this.name = name;
     this.encodedName = firebaseEncode(this.name);
-    this.failureMessage = 'TODO ERROR MESSAGE, (e.g. stackoverflow error line 13)';
+    this.failureMessage = 'Successful';
```
bcoe, line 23: Nit: I find `failureMessage` being `Successful` a bit weird, should this be something like `statusMessage`.

I'm okay with avoiding the refactor and keeping `failureMessage` depending on the hassle ... I almost like this little quirk 😆

packages/api/test/post-build-test.js
```diff
@@ -54,6 +61,47 @@ describe('Posting Builds', () => {
...
+    assert.strictEqual(testcases.length, 6);
+    const sols = [true, true, false, true, true, false];
+    for (let i = 0; i < testcases.length; i++) {
+      assert.strictEqual(testcases[i].successful, sols[i]);
+    }
+    // failure messages will be mangled which is inevitable when taking this approach
+    // but good that core functionality still works
+  });
+
+  it('should handle a test with a not flat structure and repeated tests', async () => {
```
bcoe, line 95: nit: I would probably call this a `nested structure`, just  terminology wise.

packages/api/test/res/mangledtap.tap
```diff
@@ -0,0 +1,20 @@
...
+EMPTYLINE
+
+ok 1-> Testing Box duplicatetest
+ok 2 Testing Box should assert volume of the box to be 6000
+not ok 3 Testing Box should throw an error
+
+
+  AssertionError: expected 1 to equal 2
+      at Context.<anonymous> (test/Box.test.js:20:12)
+      at processImmediate (internal/timers.js:456:21)
```
bcoe, line 11: I like  that  we're building a  corpus of tap output for tests 👍

PR #84, opened 2020-07-16
feat: tweaks for org page and repo page

StuartRucker:
Many small changes that were needed to get the org page and repo page all the information they need

- Store more information for tests including the name, whether it passed on their most recent run
- Maintain a counter using the `firestore.fieldvalue.increment` function of the lifetime number of passes and fails of a test case. This method required an additional check to make sure the `addBuild` function was not called twice on the same build.
- Store the time of the last update for repositories and tests
- Break up the routes needed for a repo page into three separate routes for metadata `/repo/:org/:repo`, builds `/repo/:org/:repo/builds` and tests `/repo/:org/:repo/tests`
- The new `/repo/:org/:repo/tests` endpoint sorts the tests in order of failing, flaky, normal, and then secondarily on the date of the last run. To accomplish this search efficiently, and additional field of `searchindex` was added to tests that is 2 if failing, 1 if flaky, and 0 otherwise.
- Renamed the `buildhandler` and `repoorghandler` to `repohandler` and `orghandler` respectively, so that these handlers would reflect the pages that they handle requests for.
- Marked 20% of the fake tests as failing in the `scripts/populate.js` file so that the mock data would be more useful for testing

The JSON for these two pages should resemble: 

For org page:
So the four routes used and the json are 
/repo?org=node https://gist.github.com/StuartRucker/411a1094895a70dd29937bf06f8f402a

For repo page:
/repo/:org/:repo  https://gist.github.com/StuartRucker/107f338dbf80bf87682b7a90dac397df
/repo/:org/:repo/builds https://gist.github.com/StuartRucker/c36a6995e3e7bfcaf7d87cc32e71fce1
/repo/:org/:repo/tests https://gist.github.com/StuartRucker/c7a41aac6edad864c3d038a0dfef41c0

### DIFF
--- a/packages/api/src/get-repo.js
+++ b/packages/api/src/get-repo.js
@@ -31,11 +31,7 @@ class GetRepoHandler {
       try {
         const repoid = firebaseEncode(req.params.orgname + '/' + req.params.reponame);
 
-        const d = new Date();
         const metadata = await this.client.collection(global.headCollection).doc(repoid).get();
-        const d2 = new Date();
-        const totalTime = d2.getTime() - d.getTime();
-        console.log('this took ' + totalTime);
 
         if (metadata.data()) {
           res.send(metadata.data());
--- a/packages/api/src/get-repo.js
+++ b/packages/api/src/get-repo.js
@@ -20,15 +20,35 @@ const { InvalidParameterError, handleError } = require('../lib/errors');
 const { isJson } = require('../util/validation');
 const FILTERS_ALLOWED = ['tag', 'ref', 'os', 'limit', 'matrix'];
 
-class GetBuildHandler {
+class GetRepoHandler {
   constructor (app, client) {
     this.app = app;
     this.client = client;
   }
 
   listen () {
-    // return the recent builds based on parameters
     this.app.get('/api/repo/:orgname/:reponame', async (req, res, next) => {
+      try {
+        const repoid = firebaseEncode(req.params.orgname + '/' + req.params.reponame);
+
+        const d = new Date();
+        const metadata = await this.client.collection(global.headCollection).doc(repoid).get();
+        const d2 = new Date();
+        const totalTime = d2.getTime() - d.getTime();
+        console.log('this took ' + totalTime);
+
+        if (metadata.data()) {
+          res.send(metadata.data());
+        } else {
+          res.status(404).send({ error: 'Did not find repo' });
+        }
+      } catch (err) {
+        handleError(res, err);
+      }
+    });
+
+    // return the recent builds based on parameters
+    this.app.get('/api/repo/:orgname/:reponame/builds', async (req, res, next) => {
       try {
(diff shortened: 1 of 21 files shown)


packages/api/src/get-repo.js
```diff
@@ -20,15 +20,35 @@ const { InvalidParameterError, handleError } = require('../lib/errors');
...
-    // return the recent builds based on parameters
     this.app.get('/api/repo/:orgname/:reponame', async (req, res, next) => {
+      try {
+        const repoid = firebaseEncode(req.params.orgname + '/' + req.params.reponame);
+
+        const d = new Date();
+        const metadata = await this.client.collection(global.headCollection).doc(repoid).get();
+        const d2 = new Date();
+        const totalTime = d2.getTime() - d.getTime();
+        console.log('this took ' + totalTime);
```
bcoe, line 38: I would make these timing messages a bit more descriptive, for the benefit of our server logs, one option perhaps  look at using middleware like [response-time](http://expressjs.com/en/resources/middleware/response-time.html) and we'd get timings for all requests.

packages/api/src/get-test.js
```diff
@@ -78,6 +78,23 @@ class GetTestHandler {
         handleError(res, err);
       }
     });
+
+    // returns all tests that failed or are flaky, and then any remaining tests
+    this.app.get('/api/repo/:orgname/:reponame/tests', async (req, res, next) => {
+      try {
+        const repoid = firebaseEncode(req.params.orgname + '/' + req.params.reponame);
+        const limit = 30;
```
bcoe, line 86: I would add a comment that notes we should make result size configurable.

packages/api/src/get-repo.js
StuartRucker: deleted. Left over from testing.

packages/api/src/get-test.js
StuartRucker: Added limit query param

packages/frontend/package.json
```diff
@@ -50,7 +50,7 @@
     "karma-jasmine": "~3.0.1",
     "karma-jasmine-html-reporter": "^1.4.2",
     "protractor": "~7.0.0",
-    "puppeteer": "^3.3.0",
+    "puppeteer": "latest",
```
bcoe, line 53: this is a bit weird, can we revert this dependency update?  If we do want to upgrade puppeteer, we can do:

```
npm i puppeteer@latest --save-dev
```

But this should be in another PR>.

packages/api/test/addbuild-getbuild.js
```diff
@@ -305,6 +308,24 @@ describe('Add-Build', () => {
       assert('error' in ansObj);
       assert.strictEqual(resp.status, 404);
     });
+
+    it('Can returns all tests in the right order', async () => {
```
bcoe, line 312: just a grammar nit:

```
it returns all test cases   in the right order
```

PR #95, opened 2020-07-18
feat: added the heat map component

gis727:
Added [D3.js](https://d3js.org/)
Added the component (empty).


packages/frontend/package.json
```diff
@@ -28,6 +28,8 @@
     "@angular/platform-browser": "~9.1.9",
     "@angular/platform-browser-dynamic": "~9.1.9",
     "@angular/router": "~9.1.9",
+    "@types/d3": "^5.7.2",
+    "d3": "^5.16.0",
```
bcoe, line 32: It's cool that you chose `d3` for the chart library, I always loved the infographics that NYT would create using it.

PR #96, opened 2020-07-18
feat: added a batch service for the heat map

gis727:
Note: Requires #95/#114
The service transforms an array of builds into an array of batches.
Each batch contains every build run on a specific day.


packages/frontend/src/app/heat-map/services/batch/batch.service.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      const batches: BuildBatch[] = service.buildBatches(builds);
+
+      expect(batches.length).toEqual(1);
+      const batch = batches[0];
+      expect(batch.builds.length).toEqual(1);
+      expect(batch.builds[0]).toEqual(jasmine.objectContaining(builds[0]));
+    });
+
+    it('should return every build ran the same day in the same batch', () => {
```
bcoe, line 44: grammar nit: "should return every build run the same day in the same batch".

gis727:
Yes ! @bcoe #95 adds the stub, #96 and #97 add the needed services and #98 pulls everything together.

PR #97, opened 2020-07-18
feat: added the heat map service

gis727:
Note: requires #96.
This service handles the drawing for the heatmap (with [D3.js](https://d3js.org/))


packages/frontend/src/app/heat-map/services/heat-map/heat-map.service.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  public draw(
+    config: Congifuration,
+    batches: BuildBatch[],
+    selector: string
+  ): EventEmitter<BuildBatch> {
+    this.clearMap();
+    this.setConfig(config, batches, selector);
+    return this.setSquares();
+  }
```
gis727, line 50: Entry point

PR #98, opened 2020-07-18
feat: integrated the heat map

gis727:
Note: requires #97.

Added the heat map template and integrated the heat map in the repository component.

[Deployed](https://flaky-dev-staging.web.app/org/nodejs/node) at the time of writing.


packages/frontend/src/app/heat-map/heat-map.component.css
```diff
@@ -13,3 +13,71 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+.flaky {
+    background-color: #fff4e0;
+}
+.failing {
+    background-color: #ffdede;
+}
+.success {
+    background-color: greenyellow;
+}
```
gis727, line 83: [flex-layout](https://github.com/angular/flex-layout) will be integrated in a different PR to get rid of some of this css.

packages/frontend/src/app/heat-map/heat-map.component.html
```diff
@@ -13,3 +13,35 @@
...
+      <mat-nav-list>
+      <cdk-virtual-scroll-viewport itemSize="50" id="builds-viewport">
+          <p id="build-view-date">Builds on {{selectedBatchDate}}</p>
+          <mat-divider></mat-divider>
+          <div
+          *cdkVirtualFor="let build of selectedBuilds; "
+          class="build"
+          [class.flaky]="!build.failcount && build.flaky"
+          [class.failing]="build.failcount"
+          [class.success]="!build.failcount && !build.flaky">
```
chingor13, line 31: In general, I'm not a fan of having too much logic in templates.

PR #99, opened 2020-07-19
feat: updated the main page

gis727:
Removed the organisation name from the repositories list (since they are all in the same organisation).
Linked the **last update on ...** to the actual data (instead of the mock).
Added the auxiliary click support for the repositories.
Added an indicator for passing tests count.
Added a progress spinner that hides when the repositories are received.
Added a text when there is no repository.


packages/frontend/src/app/main/main.component.ts
```diff
@@ -33,29 +34,37 @@ export class MainComponent implements OnInit {
...
-    });
+    const search = {
+      query: repo,
+      filters: foundParams.filters,
+    };
+
+    if (org) search.filters.push({name: 'org', value: org});
+
+    return search;
   }
```
gis727, line 68: Extracted the search params extraction in **_getSearch_**

packages/frontend/src/app/main/repo-list/repo-list.component.html
```diff
@@ -52,7 +51,7 @@
                     <!-- LAST RUN TIME -->
                     <td class="last-update">
                         <span class="last-update-label">last update on&nbsp;</span>
-                        <span class="last-update-date">May 19, 2020</span>
+                        <span class="last-update-date">{{getLastUpdate(repo)}}</span>
```
gis727, line 54: Using the real data now.

packages/frontend/src/app/main/repo-list/repo-list.component.html
```diff
@@ -19,13 +19,12 @@
         class="repo"
         [class.flaky-repo]="!repo.numfails && repo.flaky"
         [class.failing-repo]="repo.numfails"
-        (click)="onElementClick(routes.repo.link(repo.organization, repo.name), $event)">
+        (click)="onElementClick(routes.repo.link(repo.organization, repo.name), $event)"
+        (auxclick)="onElementClick(routes.repo.link(repo.organization, repo.name), $event, true)">
```
gis727, line 23: When the repository is clicked with the wheel, it now opens a new tab.

PR #101, opened 2020-07-20
feat: privacy supplements

cedpeters:
This PR:
1. Adds functionality to delete expired sessions
2. Adds a cron job which runs that functionality every five minutes
3. Protects any endpoint starting with '/protected/' so that only authenticated sessions will receive data back from them
4. Doesn't store usernames in a session unless the user is an admin


packages/api/package.json
```diff
@@ -28,17 +28,24 @@
...
+    "sinon": "^9.0.2",
+    "supertest": "^4.0.2",
     "tap-parser": "^10.0.1",
+    "url": "^0.11.0",
     "uuid": "^8.2.0"
   },
   "devDependencies": {
     "c8": "^7.1.2",
     "mocha": "^7.2.0",
-    "nock": "^13.0.2",
```
StuartRucker, line 41: Is there a reason this was moved from dev dependency. Also chai.

packages/api/src/repository.js
```diff
@@ -78,6 +78,69 @@ class Repository {
...
+        }
+      } catch (err) {
+        await batch.delete(doc.ref);
+        console.log('DELETING');
+        numSeen++;
+      }
+      if (numSeen >= 300) {
+        console.log('COMMITTING');
+        await batch.commit();
+        batch = client.batch();
```
StuartRucker, line 138: I cant quite tell from looking at it, but is it ever possible you would exceed the 500 limit for batch operations?

cedpeters: I had some trouble with this bit of code, it does seem to struggle when there are a lot of sessions to delete. It works for small amounts, which is all that's needed in production since it will be constantly running, so the amount of sessions will never get too big, but I plan to return to this later and test/fix it up more.

packages/api/server.js
```diff
@@ -49,8 +61,8 @@ app.use(
   })
 );
 
-app.get('/api/repos', async (req, res) => {
-  const repository = new Repository(null);
+app.get('/protected/api/repos', async (req, res) => {
```
bcoe, line 64: Things are far enough along at this point, that I think we can simplify things by only protecting a few specific routes, like the routes we're adding that support deleting repositories and tests.

We shouldn't need to adda new prefix like this, I think instead we could do something like this:

```
app.get('/api/repos', isLoggedIn.isLoggedIn, async (req, res) => {
  ....
})
```

Each route can have middleware optionally added as the next parameter, it then works like a chain where one need to finish before the next.

packages/api/src/isLoggedIn.js
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const moment = require('moment');
+
+module.exports.isLoggedIn = (req, res, next) => {
+  if (req.session && req.session.expires != null && moment().isBefore(moment(req.session.expires))) {
```
bcoe, line 18: this looks really solid 👍

packages/api/test/logged-in-test.js
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const assert = require('assert');
+const fetch = require('node-fetch');
+const { describe, it } = require('mocha');
+
+describe('flaky express server', () => {
+  it('rejects unauthorized session', async () => {
+    const resp = await fetch('http://0.0.0.0:3000/protected/api/repos');
```
bcoe, line 21: I think we probably don't want to add a `protectewd/` prefix to `/api/repos`, better that we test a new route that's protected with admin settings, a good option might be picking one of the deletion routes that @StuartRucker has added in.

packages/api/src/repository.js
sofisl: I wonder if maybe you could get an array for all the docs, and instead of sifting one-by-one, instead using .filter() to get a new array of expired docs. THen you could split these up into the appropriate groups (500?) and then delete those. 

As a further note, I would recommend against doing the same thing in a catch block as your try block. It looks like you're trying to build a retry but that is not the preferred way of doing that. For now, I would just catch the error, and somehow make sure you're putting that doc back in a list to look over, rather than immediately trying to delete again.

packages/api/test/server-test.js
```diff
@@ -53,12 +102,35 @@ describe('flaky express server', () => {
...
       assert.strictEqual(resp.url, frontendUrl);
     });
-  });
 
-  describe('/api/session', async () => {
-    it('returns 200 state when request session info', async () => {
-      const resp = await fetch('http://0.0.0.0:3000/api/session');
-      assert.strictEqual(resp.status, 200);
+    // Second major issue: test auth
+    it.skip('successfully does auth process', async () => {
```
sofisl, line 107: make sure to delete these test controls! (only, skip, etc).

bcoe:
@cedpeters converted this to draft rather than closing, so that it's handy for reference.

cedpeters:
> @cedpeters converted this to draft rather than closing, so that it's handy for reference.

Thanks, @bcoe!

bcoe:
I think we've moved this work at this point into #124, let me know if I'm wrong.

cedpeters:
Yep, I copied most of what I need over. Let's just not delete the branch so I can reference it if I need it, but definitely don't need this PR anymore.

PR #102, opened 2020-07-20
feat: do not allow adding of private repos

StuartRucker:
Added a small check to make sure the repository from which data is being added isnt private


packages/api/src/post-build.js
```diff
@@ -180,6 +180,10 @@ class PostBuildHandler {
           throw new UnauthorizedError('Must have valid Github Token to post build');
         }
 
+        if (req.body.metadata.github.event.repository.private) {
+          throw new UnauthorizedError('Flaky does not store tests for private repos');
```
bcoe, line 184: One thought I have from a security and privacy perspective, it might make sense for us to just send the select fields that we care about from our action, rather than posting the entire payload.

If we were to to this, we could perhaps cleanup the payload a bit to be:

```
{
  meta: {
    private: true
  }
}
```

Keep in mind we will eventually want to support  systems other than GitHub, and this ties us pretty tightly into the action specific schema.

PR #110, opened 2020-07-21
feat: can order repos in org

StuartRucker:
Added `orderby` query parameter to `/api/org/:org` endpoint that takes values `priority` `activity` and `name`. Priority first surfaces repos based on the number of failures in the last build, and secondarily based on the number of flaky tests.

Doing this required significant database restructuring in order to add a compound query index for org name and these orderby fields across all databases (testing, staging, etc). Thus, all data is now nested in a collection titled repos.

Some of the database routes would be 
`testing/main/repos`
`testing/analytics-testsuite-353234/repos`
`staging/main/repos`

Note that they all end in repos.

@madelyndubuk will need to change the environment variables.

Additionally, added `offset` query parameter to the `/api/repo/:org/:repo/tests` and `/api/org/:org`


packages/api/src/get-org.js
```diff
@@ -90,7 +90,23 @@ class GetOrgHandler {
         if (!req.params.orgname) {
           throw new InvalidParameterError('requires org parameter');
         }
-        const snapshot = await this.client.collection(global.headCollection).where('lower.organization', '==', req.params.orgname.toLowerCase()).get();
+        const limit = parseInt(req.query.limit || 30);
```
gis727, line 93: @bcoe , it think this is where the 30 comes from.

PR #118, opened 2020-07-22
feat: added the 'orderby' filter in the main page

### LINKED ISSUE #117: add the 'order by' filter in the main page
The options should be `priority | name | activity`.

The options should appear only when:
- [x] The repositories data are received (page not loading).
- [x] At least 2 repositories were found.
- [x] The found repositories do not come from a search.

gis727:
Fixes #117 and requires #116.


packages/frontend/src/app/main/main.component.spec.ts
```diff
@@ -126,4 +139,82 @@ describe('MainComponent', () => {
...
+    component.repoName = '';
+    fixture.detectChanges();
+    tick();
+
+    const filterComponent = fixture.debugElement.query(By.css('app-filters'));
+
+    expect(filterComponent).toBeNull();
+  }));
+
+  it('should hide the filters when a repository is being searched', fakeAsync(() => {
```
gis727, line 179: @bcoe , @StuartRucker , this is the case where the displayed repositories come from a search.

PR #123, opened 2020-07-22
feat: hid the paginators extra labels and controls

gis727:
The paginators now only have the "previous" and "next" arrows.

Screenshot 2020-07-22 at 1 25 15 PM


bcoe:
I like it 👍

PR #124, opened 2020-07-22
feat: permissions auth dance

cedpeters:
All use of sessions for authentication in the API has been removed. The /session endpoint has been deleted. The /auth endpoint has been changed from a GET to a POST, and it is supposed to receive a JSON which will say what action the user has requested (e.g. delete a specific test). At the end of the auth dance, the API will now perform the requested action (if permitted), rather than save session information.

The actual storage and performance of the action is left for a future PR. This is mainly just ripping out the sessions and stubbing out future functionality.


cedpeters:
Functionality to send state to Github, do the auth dance, and verify that the state is correct at the end of the dance. 90%+ test coverage, though the specific values posted to Github are not being tested. The general back-and-forth with Github is tested with Sinon, though.

packages/api/server.js
```diff
@@ -61,75 +45,54 @@ app.get('/api/repos', async (req, res) => {
...
-  // TODO allow the requester to give search/filter criterion!
+
   res
     .status(200)
     .send(jsonObject)
     .end();
 });
 
-app.get('/api/auth', (req, res) => {
+app.post('/api/repo/:orgname/:reponame/test', async (req, res) => {
```
cedpeters, line 55: I tried making it DELETE as we discussed, but found it much easier to redirect from a POST endpoint than from DELETE.

packages/api/server.js
```diff
@@ -37,21 +33,9 @@ global.headCollection = process.env.HEAD_COLLECTION || 'testing-buildsget';
...
-    }),
-    secret: process.env.SESSION_SECRET || 'no secret',
-    resave: false,
-    saveUninitialized: true
-  })
-);
 
 app.get('/api/repos', async (req, res) => {
-  const repository = new Repository(null);
-  const result = await repository.getCollection('dummy-repositories');
```
bcoe, line 54: I don't think we use this `/api/repos` route for anything at this point, and could probably just delete the whole thing?

@StuartRucker could you  confirm for me

packages/api/server.js
```diff
@@ -61,75 +45,54 @@ app.get('/api/repos', async (req, res) => {
...
+
+  await repo.storeTicket({
+    action: 'delete-test',
+    repoId: repoId,
+    testname: testname,
+    state: state
+  });
+
+  const url = 'http://github.com/login/oauth/authorize?client_id=' + process.env.CLIENT_ID + '&state=' + state + '&allow_signup=false';
+  res.status(307).redirect(url);
```
bcoe, line 68: I think a `302` is usually used for OAuth 2.0 redirects.

packages/api/server.js
```diff
@@ -61,75 +45,54 @@ app.get('/api/repos', async (req, res) => {
...
-  }
+  const userData = await auth.retrieveUserData(queryObject.access_token);
 
-  req.session.user = resultJSON.login;
-  const repository = new Repository();
-  const permitted = await repository.mayAccess('github', resultJSON.login);
-  if (permitted) {
-    req.session.expires = moment().add(4, 'hours').format();
+  // todo send a message to the frontend to put a banner on the page indicating whether the action has been performed
+  if (repo.performTicketIfCan(userData)) {
```
bcoe, line 89: nit: but I might go with `completeTicketIfAllowed`, or `performTicketIfAllowed`.

packages/api/server.js
bcoe: I think `DELETE` is probably the correct verb for this type of operation, the basic idea of REST is this:

* `GET`: I'm fetching something.
* `POST`: Create something.
* `PUT`: update something.
* `DELETE`: I intend to delete something.

I did a little test, and this should work:

If you switch `app.post('/api/repo/:orgname/:reponame/test', async (req, res) => {` to
`app.delete('/api/repo/:orgname/:reponame/test', async (req, res) => {`.

```js
    it.only('generates a GitHub redirect', async () => { 
      stubs.push(sinon.stub(repo, 'storeTicket').returns(true));
      const resp = await fetch('http://0.0.0.0:3000/api/repo/my-org/my-repo/test', {
        redirect: 'manual',
        method: 'DELETE'
      });
      assert(resp.headers.get('location').includes('github.com/login/oauth'));
    })
```

I think another thing we should probably change, is that for the route that deletes a specific test we'll want to do:

```
/api/repo/:orgname/:reponame/test/:testid
```

since we'll need to know the test ID too 👍

packages/api/server.js
```diff
@@ -61,75 +45,54 @@ app.get('/api/repos', async (req, res) => {
...
+  const repoId = req.params.reponame;
+  const testname = req.query.name;
   const state = v4();
-  req.session.authState = state;
-  const url = 'http://github.com/login/oauth/authorize?client_id=' + process.env.CLIENT_ID + '&state=' + req.session.authState + '&allow_signup=false';
-  res.status(302).redirect(url);
+
+  await repo.storeTicket({
+    action: 'delete-test',
+    repoId: repoId,
```
bcoe, line 62: I think we'd be smart to store a `redirect` with the ticket too, and then the folks calling the DELETE endpoint can specify the redirect they want to land on after the delete, this way we can take the user back to the page that the initiating  the deletion on ... this is useful if you're deleting multiple tests.

packages/api/src/repository.js
```diff
@@ -50,28 +49,22 @@ class Repository {
...
-      }
-      if (moment().isBefore(moment(expiration))) {
-        solution.permitted = true;
-        solution.expiration = moment(expiration).format();
-        solution.login = data.user;
-      }
-    }
-    return solution;
+  async storeTicket (ticketToPerform) {
+    await this.createDoc('tickets/ticket-to-perform', {
```
bcoe, line 53: I believe this will keep overwriting a document named `tickets/ticket-to-perform`. Instead of doing this, I would just use the same unique identifier you've already created for `state` and create a document:

```
ticets/${state}
```

I think we should consider getting all this database work hammered out in a second PR after we land this though.

packages/api/test/auth-test.js
```diff
@@ -0,0 +1,44 @@
...
+      const stubbed = sinon.stub(fetch, 'Promise').returns(Promise.resolve(data));
+
+      const string = await auth.retrieveAccessToken('code', 'ticketState');
+      assert.strictEqual(string, null);
+      stubbed.restore();
+    });
+  });
+
+  describe('retrieveUserData', async () => {
+    it('returns null if the result status is not 100', async () => {
```
bcoe, line 35: should this be `200` not `100`?

packages/api/test/post-build-test.js
```diff
@@ -34,14 +34,14 @@ const { deleteRepo } = require('../lib/deleter');
...
 
-nock.disableNetConnect();
-nock.enableNetConnect(/^(?!.*github\.com).*$/); // only disable requests on github.com
-
 const assert = require('assert');
 const fetch = require('node-fetch');
 
 describe('Posting Builds', () => {
   before(async () => {
+    nock.disableNetConnect();
```
bcoe, line 42: it seems a bit weird that we  moved this logic down into the `before` statement... I find this a logic a little weird in general.

@StuartRucker why does it `enableNetConnect` for GitHub.com, but in the comment say it's disabling it? 

Ideally we shouldn't be making an external connections in tests.

packages/api/test/repository-test.js
```diff
@@ -71,19 +71,18 @@ describe('Repository', () => {
...
+    it('performs requested test delete when user has write permission to repo', async () => {
 
-    it('returns not permitted if there is no expiration date', async () => {
-      await repo.createDoc('express-sessions/test-session', {});
-      const permission = await repo.sessionPermissions('test-session');
-      assert.deepStrictEqual(permission, { permitted: false, expiration: null, login: null });
-      await repo.deleteDoc('express-sessions/test-session');
-    });
+      /**
+    Sample data for deleting a test:
```
bcoe, line 78: Let's get rid of commented out code.

packages/api/test/server-test.js
```diff
@@ -12,18 +12,27 @@
...
 
 const assert = require('assert');
 const fetch = require('node-fetch');
+const sinon = require('sinon');
+const querystring = require('querystring');
+const auth = require('../src/auth.js');
+const repo = require('../src/repository.js');
+// const nock = require('nock');
+
+// nock('http://github.com')
```
bcoe, line 26: likewise, I think we shouldn't need to use nock for the new tests that are added, let's just get rid of the commented out code.

packages/api/test/server-test.js
```diff
@@ -37,16 +46,47 @@ describe('flaky express server', () => {
...
+    it('redirects to Github', async () => {
+      const resp = await fetch('http://0.0.0.0:3000/api/repo/fakeorg/fakerepo/test?name=dummytest', {
+        headers: { redirect: 'manual' },
+        method: 'post'
       });
-      assert.strictEqual(resp.url.includes('github.com/login'), true);
+
+      assert.deepStrictEqual(resp.url.includes('github.com/login'), true);
     });
   });
```
bcoe, line 58: Mind adding a test for generating the redirect too? something like this:

```js
    it.only('generates a GitHub redirect', async () => { 
      stubs.push(sinon.stub(repo, 'storeTicket').returns(true));
      const resp = await fetch('http://0.0.0.0:3000/api/repo/my-org/my-repo/test', {
        redirect: 'manual',
        method: 'DELETE'
      });
      assert(resp.headers.get('location').includes('github.com/login/oauth'));
    })
```

You could also make this test validate that the appropriate data from the request is being stored in the ticket (this would be a good addition to the test).

packages/api/server.js
cedpeters: Yep this was a demo endpoint back when we were getting started. Stuart can confirm, but I'm fairly sure we don't use it for anything.

packages/api/test/post-build-test.js
cedpeters: I'm not currently using nock in server-test, but the test for DELETE /repo, which does a manual redirect, was throwing an error because nock was interfering with any github calls (even though that test wasn't actually hitting the github endpoint, since it was doing a manual redirect). The test passed when run in isolation, but failed when run alongside post-build-test. Moving that logic into the before statement fixed the issue.

packages/api/server.js
StuartRucker: Yeah it was replaced with /org/:org

packages/api/test/post-build-test.js
StuartRucker:
```
nock.disableNetConnect();
nock.enableNetConnect(/^(?!.*github\.com).*$/); // only disable requests on github.com
```
Was the only way to allow all requests EXCEPT github.com. Basically disable everything, and then enable anything matching a pattern that is NOT github

packages/api/server.js
```diff
@@ -38,99 +33,60 @@ global.headCollection = process.env.HEAD_COLLECTION || 'testing/main/repos';
...
-  const result = await fetch('https://api.github.com/user', {
-    method: 'get',
-    headers: { 'content-type': 'application/json', 'User-Agent': 'flaky.dev', Authorization: 'token ' + queryObject.access_token }
-  });
-
-  const resultJSON = await result.json();
+  const queryObject = await auth.retrieveAccessToken(req.param('code'), storedTicketState);
 
-  if (result.status !== 200) {
+  if (queryObject == null) {
```
bcoe, line 75: nit: tiny non blocking nit, I prefer `===` so that we're 100% sure of the variable's type.

PR #125, opened 2020-07-22
fix: description not required field

StuartRucker:
Does not require a github repository to have a description to post a build. Additionally does not require OS or Ref. Necessary for the yargs repository to be used for testing.


StuartRucker:
> Is there a test to update?

Added test

PR #129, opened 2020-07-23
feat: test details expansion panel

madelyndubuk:
- added expansion panel for test details on test list
- can open more than one panel at once
Screenshot 2020-07-22 at 2 28 38 PM


bcoe:
@madelyndubuk I like how this is looking from a design perspective 👌 please don't hesitate to reach out when you want a more thorough review.

bcoe:
A few UX nits/questions:

## 1. When in the minimized view, it feels like there is a bit too much left and right padding:

Screen Shot 2020-07-24 at 11 50 43 AM

I think it would be fine to stick with the padding we have today, which looks sharp:

Screen Shot 2020-07-24 at 11 53 09 AM

## 2. When maximized it says "passes 0% of the time" in green ... which kind of makes me feeling like this is a good thing 😆 

I would probably make this yellow or red, if it's below some threshold like 90% passing? 

---

I think we should just make `2.` a tracking ticket, for improving that copy.

packages/frontend/src/app/repository/tests-list/tests-list.component.spec.ts
```diff
@@ -115,23 +123,18 @@ describe('TestsListComponent', () => {
     expect(component.pageIndex).toEqual(expectedPageIndex);
   });
 
-  it('should open the tests details when user clicks on a test', async () => {
+  fit('should expand the tests details when user clicks on a test', fakeAsync(() => {
```
gis727, line 126: Revert to `it` ?

packages/frontend/src/app/home/home.component.spec.ts
```diff
@@ -65,20 +65,4 @@ describe('HomeComponent', () => {
...
-    licenseButton.click();
-    tick();
-    expect(component.openLicenseDialog).toHaveBeenCalled();
-  }));
-
-  it('should call the dialog open() function when the license button is clicked', fakeAsync(() => {
-    dialogSpy = spyOn(TestBed.get(MatDialog), 'open').and.callThrough();
-    component.openLicenseDialog();
-    expect(dialogSpy).toHaveBeenCalled();
-  }));
```
gis727, line 83: Are we dropping those tests ?

packages/frontend/src/app/repository/tests-list/tests-list.component.spec.ts
madelyndubuk: oh yes thank you

packages/frontend/src/app/home/home.component.spec.ts
madelyndubuk: I think these were duplicates so I just deleted one set of them

packages/frontend/src/app/repository/tests-list/tests-list.component.html
```diff
@@ -15,49 +15,62 @@
 -->
 
 <div id="tests-container">
+    
+    <mat-accordion multi>
+        <mat-expansion-panel *ngFor="let test of renderedElements"
```
bcoe, line 20: I love that we're able to use these off the shelf components, and get such delightful behavior.

bcoe:
This is looking awesome @madelyndubuk, only nit: is wondering if it's possible to test the state change?

madelyndubuk:
Fixes #147

PR #141, opened 2020-07-24
feat(flaky-tests): add flaky.dev for flaky test tracking

bcoe:
This PR:

* adds a `tap` test runner to both the frontend and and API.
* updates our GiitHub action to upload to the flaky service.

Let's 🐕 🥫 flaky.dev.


.github/workflows/ci.yaml
```diff
@@ -27,6 +27,14 @@ jobs:
         working-directory: ./packages/api
       - run: npm test
         working-directory: ./packages/api
+      - name: track flaky tests
+        uses: GoogleCloudPlatform/flaky-service/packages/action@ee9cc51f3b42fc8e9e43bee8a2cf86d8bc82c03e
```
bcoe, line 31: @StuartRucker if we added a feature to [release-please](https://github.com/googleapis/release-please) that allowed it to publish from directories, we could automatically publish your GitHub action, so that we don't need this ugly `sha`.

GitHub actions cache, so you don't want to use a `branch` name.

bcoe:
@StuartRucker looks like we got a 200 uploading frontend and api builds, but I only see the API builds [here](https://flaky-dashboard.web.app/org/GoogleCloudPlatform/flaky-service) any thoughts on what happened?

PR #149, opened 2020-07-24
fix: can add multiple builds with the same id

StuartRucker:
When adding multiple builds with the same id, the second id has a random uuid appended to it. This should fix the issue with uploading flaky.dev's tests.

Add new fields in runs and builds `buildIdReal` which is the build id with the modification, and kept the `buildId` field raw.


packages/api/src/add-build.js
```diff
@@ -46,7 +47,7 @@ async function addBuild (testCases, buildInfo, client, collectionName = 'reposit
   // if this build has already been posted, skip
   const thisBuildExists = await dbRepo.collection('builds').doc(buildInfo.buildId).get();
   if (thisBuildExists.exists) {
-    return;
+    buildInfo.buildId = buildInfo.buildId.toString() + uuidv4();
```
bcoe, line 50: will this show up in the UX in places where we show the workflow ID?

Screen Shot 2020-07-24 at 5 07 44 PM

What if we made all `buildId`s `uuidv4()`, and then store the build ID reported by the test runner separately?

StuartRucker: Now stores both, with raw build id in `buildIdReal` and buildid + uuid in `buildIdReal`

tbpg: I think I'd prefer the "different" name be the one with the UUID. Something like `internalBuildId`, `buildIdInternal`, or `buildIdWithUuid`. I had to look a few times to understand/see `buildIdReal`.

packages/api/test/addbuild-getbuild.js
```diff
@@ -136,6 +136,7 @@ describe('Add-Build', () => {
 
       // ensure builds were uploaded correctly
       const builds = await client.collection(global.headCollection).doc(buildInfo[1].repoId).collection('builds').doc(buildInfo[1].buildId).get();
+      console.log('BUIDLS>DATA: ' + builds.data().percentpassing);
```
bcoe, line 139: is this a useful log message to have in for now? or  accidentally left. Happy to leave the log line if it's helping during debugging.

PR #150, opened 2020-07-24
feat: store error message of most recent failure in test document

StuartRucker:
Test document now has field `failuremessageiffailing` that has the error message for a test if `passed` is not true. This refers to the chronologically most recent build.


bcoe:
@StuartRucker nice, I'm excited to see this in action.

PR #157, opened 2020-07-26
feat: added dynamic pagination

### LINKED ISSUE #126: add pagination support in the frontend
The frontend currently fetches all repositories/tests and handles pagination locally.
Since the server is getting the support for paginations in #110 and #139, the frontend should follow.

gis727:
Fixes #126. [Deployed](https://flaky-dev-staging.web.app/org/alphaorg).

This PR:
- adds the pagination mechanism (working with the API) to the organization and the repository pages
- links the filters to the displayed results. This means that:
  - on the test page, the filters are sent to the api while fetching tests for the tests list and builds for the heatmap.
- includes #158.

Note: The size of this PR mostly comes from the new tests added for the pagination mechanism.


packages/frontend/src/app/main/repo-list/repo-list.component.spec.ts
```diff
@@ -58,40 +72,25 @@ describe('RepoListComponent', () => {
...
-
-  it('should not return to the first page when the paginator is not ready', () => {
-    const expectedPageIndex = 1;
-    component.paginator = undefined;
-    component.pageIndex = expectedPageIndex;
-
-    component.repositories = mockRepositories;
-
-    expect(component.pageIndex).toEqual(expectedPageIndex);
-  });
```
gis727, line 96: The "first" page is no longer relevant since we are fetching each page from the API.

packages/frontend/src/app/paginated-list/paginated-list.component.spec.ts
```diff
@@ -12,123 +12,287 @@
...
   });
 
   it('should create', () => {
     expect(component).toBeTruthy();
   });
 
-  it('should update the pages on initialization', () => {
-    spyOn(component, 'updatePage');
+  describe('onPageSelectionChange', () => {
+    it('should disable the paginator when a new page is selected', fakeAsync(() => {
```
gis727, line 68: This is to prevent the user from being able to select the same "next page" multiple times.

packages/frontend/src/app/paginated-list/paginated-list.component.spec.ts
```diff
@@ -12,123 +12,287 @@
...
+        {name: 'offset', value: '0'},
+      ];
+      expect(fetcher).toHaveBeenCalledTimes(1);
+      const passedArgs = fetcher.calls.mostRecent().args as object[];
+      expectedFilters.forEach(filter =>
+        expect(passedArgs[0]).toContain(jasmine.objectContaining(filter))
+      );
+    }));
+
+    it('should restore the previous index if an error occurs while fetching data', fakeAsync(() => {
```
gis727, line 177: This is important because the user should be allowed to select the next page again if something went wrong the first time.

packages/frontend/src/app/heat-map/heat-map.component.spec.ts
```diff
@@ -54,6 +54,7 @@ describe('HeatMapComponent', () => {
     fixture.autoDetectChanges(true);
     component.config.weeksToDisplay = 2;
     component.config.daysToDisplay = 7;
+    component.init('', '');
```
bcoe, line 57: what do  these two empty strings do?

packages/frontend/src/app/services/search/search.service.spec.ts
```diff
@@ -76,53 +76,55 @@ describe('SearchService', () => {
...
+        | string
+      )[];
+      expect(reposFetcherArgs[0]).toEqual('repo');
+      expect(reposFetcherArgs[1]).toEqual('org');
+      expect(reposFetcherArgs[2]).toEqual([
+        {name: 'orderby', value: 'priority'},
+      ]);
+    }));
+
+    it("should not add the 'order by priority' filter to the search if it is already present", fakeAsync(() => {
```
bcoe, line 110: good thing to check 👍

packages/frontend/src/app/services/search/search.service.spec.ts
```diff
@@ -76,53 +76,55 @@ describe('SearchService', () => {
...
+        reponse.repos.forEach((repo, index) => {
           expect(repo.name).toEqual(repositories.repos[index].name);
         });
         done();
       });
     });
 
-    it("should add the 'order by priority' filter to the search if itsn't already present", done => {
-      const search: Search = {query: '', filters: []};
+    it("should add the 'order by priority' filter to the search if itsn't already present", fakeAsync(() => {
```
bcoe, line 89: typo `itsn't` -> `isn't`.

bcoe, line 89: do we have a test that makes sure the value of `order by priority` changes to a different `priority` value when the component is interacted with?

packages/frontend/src/app/heat-map/heat-map.component.spec.ts
gis727: If they were used in those tests, that line would be `component.init('repoName', 'orgName')`.

packages/frontend/src/app/services/search/search.service.spec.ts
gis727: Yes, in the [filters component](https://github.com/GoogleCloudPlatform/flaky-service/blob/master/packages/frontend/src/app/filters/filters.component.spec.ts#L63L76).

PR #162, opened 2020-07-27
build: add release-please for action releases

bcoe:
This adds automation that will generate release tags for our GitHub action, so that folks can use specify a version number they wish to use, e.g., `flaky-servicen@v1.6.0`.


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,15 @@
+ on:
+   push:
+     branches:
+       - main
```
StuartRucker, line 4: Should this be master for us?

bcoe: great catch @StuartRucker 👍

PR #163, opened 2020-07-28
build: always post flaky tests

bcoe:


StuartRucker:
LGTM. Same approach I needed to use on yargs https://github.com/StuartRucker/yargs-clone-testing/blob/master/.github/workflows/ci.yaml

PR #165, opened 2020-07-28
fix: delete stale test

cedpeters:


packages/api/test/repository-test.js
```diff
@@ -59,18 +59,6 @@ describe('Repository', () => {
     });
   });
 
-  describe('mayAccess', async () => {
-    it('returns false if a user is not permitted to log in via Github', async () => {
-      const permitted = await repo.mayAccess('github', 'fake');
```
bcoe, line 64: The trick we can use here as we write new tests, will be using a `uuid` rather than a static name like `github`, that way multiple suites can run in parallel too.

We might eventually then want a sweeping job that cleans up test data older than N-minutes ... _speaking of which we should make sure we have a create time on all our data._

cedpeters: This is a holdover from when we were going to have a list of admin usernames in the database, and check them. The mayAccess method checked for a username, and the 'github' argument tells it to check specifically the list of github usernames (in case we eventually let admins authenticate with other platforms).

PR #171, opened 2020-07-28
refactor: updated the repository page

gis727:
This PR:
- Adds a spinner that shows that the repository page is loading. The spinner disappears when both the heat map and the tests list are ready.
- Adds the `No tests found` text that is shown if no tests were found.
- Removes all filters from the tests list API requests because tests will not be filtered.


packages/frontend/src/app/repository/repository.component.spec.ts
```diff
@@ -177,4 +185,68 @@ describe('RepositoryComponent', () => {
 
     expect(location.path()).toEqual(RouteProvider.routes._404.link());
   }));
+
+  it('should hide the spinner when tests list is ready', fakeAsync(() => {
```
bcoe, line 189: I've said this before, but truly appreciate how thorough we're all being with tests. This will create a great foundation for future folks contributing to the project.

PR #173, opened 2020-07-28
feat: add database and Github API to auth process

cedpeters:
Fill in the repository functions which store and retrieve the ticket from the database. Also add basic permissions check using Github.


packages/api/src/auth.js
```diff
@@ -36,8 +36,8 @@ class Auth {
     return querystring.parse(respText);
   }
 
-  async retrieveUserData (accessToken) {
-    const result = await this.doFetch('https://api.github.com/user', {
+  async retrieveRepoData (accessToken) {
+    const result = await this.doFetch('https://api.github.com/user/repos', {
```
cedpeters, line 40: This is querying all repos for which the user has an explicit role: https://docs.github.com/en/rest/reference/repos#list-repositories-for-the-authenticated-user

All auth checks are based on this data. I'm a bit concerned it's not quite right, though, because when I ran it for me it didn't include flaky.dev. An alternative would be to query all data about the repo for which the action is being requested, but that requires knowing the repo's owner: https://docs.github.com/en/rest/reference/repos#check-if-a-user-is-a-repository-collaborator

bcoe: @cedpeters there's an API method that gives you the permission level on a given repository for a given user:

https://docs.github.com/en/rest/reference/repos#check-if-a-user-is-a-repository-collaborator

I think we could use this method, and confirm that the username returned via the OAuth dance.

Alternatively, we could use:

https://docs.github.com/en/rest/reference/repos#get-repository-permissions-for-a-user

Which gives us the actual permissions for a single user ... this might be better as it gives us a bit more granularity, letting us know if the user is an admin vs., a collaborator.

packages/api/src/repository.js
```diff
@@ -42,26 +42,42 @@ class Repository {
     return result;
   }
 
-  async mayAccess (platform, login) {
-    const collection = client.collection('permitted-users/' + platform + '/users').where('login', '==', login);
+  async storeTicket (ticketToPerform) {
+    return this.createDoc('tickets/' + ticketToPerform.state, {
```
bcoe, line 46: stylistic nit, I've been tending to use template strings these days:

```
`tickets/${ticketToPerform.state}`
```

not a huge difference just my personal preference.

packages/api/src/repository.js
```diff
@@ -42,26 +42,42 @@ class Repository {
...
+    return data.ticket;
   }
 
-  async storeTicket (ticketToPerform) {
+  actionIsAllowed (userRepoData, fullName, action) {
+    /**
+    Calculate whether the repository an action is being requested for
+    is on the list of the user's repositories.
+    */
+    const isOnUserRepoList = userRepoData.some(item => item.full_name === fullName);
```
bcoe, line 62: I think based on those couple methods I shared, we could simplify this into a:

```
userCanWriteRepo(user, repo)
```

packages/api/test.tap
```diff
@@ -0,0 +1,3 @@
+1..70
```
bcoe, line 1: whoops, we should delete this file, not sure how it got checked in (_I added an ignore rule to .gitignore)_.

packages/api/test/repository-test.js
```diff
@@ -59,6 +59,47 @@ describe('Repository', () => {
     });
   });
 
+  describe('ticket handlers', async () => {
+    const state = 'testing-state';
```
bcoe, line 63: to ensure that multiple runs of the test suite can happen in parallel, I would make this a `uuid` just as you would in the normal workflow.

packages/api/test/repository-test.js
```diff
@@ -59,6 +59,47 @@ describe('Repository', () => {
...
+        assert.deepStrictEqual(result, ticket);
+
+        await repo.deleteDoc('tickets/' + state);
+      });
+    });
+  });
+
+  describe('actionIsAllowed', () => {
+    it('permits test deletion when the userr is not an admin but has access to the repo', () => {
+      const res = repo.actionIsAllowed([{ full_name: 'fake-org/fake-repo' }], 'fake-org/fake-repo', 'delete-test');
```
bcoe, line 99: I think you could potentially use a similar approach here with the mocking we did last week, where you mock the step that returns the permission object from GitHub. I would test with the real thing first, and make sure we understand what responses are possible.

packages/api/src/auth.js
cedpeters: The issue with those is that the path requires the owner of the repo. For example, for repo permissions for a user, the path is: 
`/repos/{owner}/{repo}/collaborators/{username}/permission`
The UI doesn't know the owner of a repo by default, and it's not currently part of the DELETE /repo endpoint. Unless the owner is just the org?

PR #181, opened 2020-07-29
fix(filters): added support for matrix filter

gis727:
This PR:
- Adds a parser for filters that have a json format:
before:
matrixBefore
after:
matrixAfter
- Adds a conditional emission for filters (only the selected filters are now sent to the API).


packages/frontend/src/app/filters/filters.component.ts
```diff
@@ -97,11 +101,27 @@ export class FiltersComponent implements OnInit {
     return options;
   }
 
+  private tryParseOption(option: string, options: Option[]): void {
+    const optionObj = JSON.parse(option);
+    const optionKeys = Object.keys(optionObj);
```
bcoe, line 106: the edge case of `optionKeys.length === 0` would be that we have an empty object `{}` as the filter?

might be worth adding a comment.

gis727: Yep, `Object.keys({}).length  === 0`.

Issue #180, opened 2020-07-29
list of builds indicates all builds are healthy, on GitHub build indicates failure


bcoe:
## Observed Behavior

The UX for `/org/GoogleCloudPlatform/flaky-service` indicates that all builds are green:

<img width="207" alt="Screen Shot 2020-07-29 at 10 34 10 AM" src="https://user-images.githubusercontent.com/194609/88833459-2339c480-d187-11ea-860e-1f27b07292a4.png">

However, when I visit the build on GitHub `186156771`, it indicates that there were some failures:

<img width="383" alt="Screen Shot 2020-07-29 at 10 36 00 AM" src="https://user-images.githubusercontent.com/194609/88833595-5419f980-d187-11ea-8843-4ffd02df985b.png">

## Expected Behavior

There have been a couple failed builds over the past week, I was surprised we don't see  any failures in the build list.

bcoe:
@StuartRucker this might just be a request to check the database.

StuartRucker:
I believe this was a symptom of the server not being able to handle mutiple posts with the same run_id. Before this pull request https://github.com/GoogleCloudPlatform/flaky-service/pull/149 which I hypothesize fixed the issue, you will notice that there is only one build for each id. However, if you click on one of the more recent dates, there are duplicate build ids listed under a particular time.

PR #182, opened 2020-07-29
feat!: optimize uploading of data

StuartRucker:
Parallelized lots of processes that did not need to be done sequentially when uploading data. The main increase in efficiency is from uploading and updating each test in parallel, but other processes are in parallel as well. (See [diagram here](https://imgur.com/WoinuTr)).

Another optimization came from storing & maintaining in each test case a list of the 15 most recent runs of the test case. This means that there no longer needs to be 15 reads per test case. It turns out, the performance benefits of this were minimal since writing takes much more time. However, this reduces the cost of an upload to 1/2 to 1/3 what it previously was.

Previously, it took 25 seconds to upload a single build with 50 test cases. Now, it averages 27 seconds to upload a build with 3000 test cases. The time varies linearly with number of test cases, so it may have taken >10 minutes to upload all 3000 tests before.

See results: [Google Sheets Graph](https://docs.google.com/spreadsheets/d/1x0hjOG9gcXXj19CIYxdbcFRQbCGIaJCwVEo8vKQMkRo/edit?usp=sharing)

Breaking Change: Now only stores test names under a build
Breaking Change: A build's flaky test count now refers to the number of tests which are **currently** flaky and failed on that build

TODO: Test to see if using Firestore batch feature speeds it up furthur

I have yet to experience this issue with the new upload method, which previously randomly came up when calling addBuild many times at once. I do not know if it is fixed, or just less likely to occur now. 

**Resolved Connection Issues**: Many of the connection issues were fixed by queuing updates to be done 50 at a time. The remaining connection issues were likely due to a combination of bad internet on development computer. On a better computer, ~20 builds with 3000 tests were able to be updated currently.

**Resolved Concurrency Issues**: All new recent dates are added using Arrayunion, and all old dates are removed using Arrayremove when mantaining a list of the 15 most recent dates on which a particular test was run.

```
 Error: 4 DEADLINE_EXCEEDED: Deadline exceeded
    at Object.callErrorFromStatus (/home/stuartrucker/flaky-dev/packages/api/node_modules/@grpc/grpc-js/build/src/call.js:30:
```26)

### CODE REVIEW

File: packages/api/lib/analytics.js [hunk 1]
```diff
@@ -12,32 +12,73 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+const NUM_STORE = 15;
```
bcoe, line 15: love this optimization.

packages/api/lib/analytics.js
```diff
@@ -12,32 +12,73 @@
...
+    });
+    this.cachedSuccess.forEach(s => {
+      this.mergedList.push({
+        date: s,
+        passed: true
+      });
     });
-    return (total > 0) ? successCnt / parseFloat(total) : -1;
+
+    this.mergedList.sort(function (a, b) {
```
bcoe, line 52: by performing this merge in the API rather than in the database, we run the potential of losing data when two updates have happened in parallel -- this is very likely to happen in cases like flaky.dev where multiple test suites are running in parallel across a few matrices ... I wonder if this might be a potential root cause of #180.

I think denormalization is a brilliant idea, for the reasons you've outlined, but I wonder if we could move either to using firebase's [push](https://firebase.google.com/docs/database/admin/save-data#node.js_9) functionality, which would ensure that two parallel updates are both appended.

You'd then just need to add one one extra operation that deletes the oldest item, if and only if there are 15 or more.

bcoe:
given this is a breaking schema change, I believe we will want to coordinate updating our collection name with the work when we deploy.

PR #185, opened 2020-07-29
feat: ui test delete button

cedpeters:
Add the "Delete Test" button to the UI

image


packages/api/server.js
```diff
@@ -28,13 +28,13 @@ const { v4 } = require('uuid');
 
 const cors = require('cors');
 
-global.headCollection = process.env.HEAD_COLLECTION || 'testing/main/repos';
+global.headCollection = process.env.HEAD_COLLECTION || 'production/main/repos';
```
bcoe, line 31: I think it's risky to default to `production` data, if it's okay let's stick with 'testing/main/repos'

... we'll be changing this collection any ways I expect 😄

packages/api/server.js
```diff
@@ -52,7 +52,7 @@ app.delete('/api/repo/:orgname/:reponame/test/:testid', async (req, res) => {
   });
 
   const url = `http://github.com/login/oauth/authorize?client_id=${process.env.CLIENT_ID}&state=${state}&allow_signup=false&scope=repo`;
-  res.status(302).redirect(url);
+  res.status(200).send(url).end();
```
bcoe, line 55: I don't think we should need to call `.end()`.

packages/api/test.tap
```diff
@@ -1,32 +1,2 @@
-1..2
+1..75
 Example app listening at http://localhost:3000
-ok 1 Repository performTicketIfAllowed deletes test if permitted
-ok 2 Repository performTicketIfAllowed does not delete test if not permitted
-# tests 2
-# pass 2
```
bcoe, line 6: If you don't mind let's delete the `.tap` file. Would love to know how these keep sneaking in, it might be worth double checking whether this file is staged for commit next time you're pushing anything, perhaps we've screwed up the ignore rules.

packages/frontend/src/app/repository/tests-list/test-details/test-details.component.html
```diff
@@ -50,8 +50,9 @@
     <!-- FLAKINESS -->
     <span *ngIf="test.flaky" class="flaky">FLAKY</span>
     <span *ngIf="!test.flaky">NOT FLAKY</span>
+    <span>.&nbsp;</span>
 
-    <span>.</span>
+    <button mat-raised-button (click)='startDeleteTest()' id='delete-test-button' class='deleteTestButton'>Delete</button>
```
bcoe, line 55: nit: I would probably hyphenate a class name `delete-test-button`.

packages/frontend/src/app/repository/tests-list/test-details/test-details.component.spec.ts
```diff
@@ -51,4 +68,25 @@ describe('TestDetailsComponent', () => {
...
+    fixture.detectChanges();
+    //spy on the new window
+    spyOn(component.windowProvider, 'open');
+    //click the test deletion button
+    const deleteButton = fixture.debugElement.nativeElement.querySelector(
+      '#delete-test-button'
+    );
+    deleteButton.click();
+    tick();
+    expect(component.windowProvider.open).toHaveBeenCalledWith(
```
bcoe, line 87: neat, I love that the framework we have in place allows for testing behavior like this so easily.

packages/api/server.js
madelyndubuk: I agree, if you need a collection with more data, I think you can use the `staging/main/repos` instead.  That is what @gis727 and I have done

packages/frontend/src/app/repository/tests-list/test-details/test-details.component.html
gis727: Yep. Also, since there will be multiple instances of this button on the dom, i think it can't have an id (since ids are unique).
Oh, also, i'm not sure its super important, but double quotes (`"`) instead of single quotes? (for the templates only)

packages/frontend/src/app/repository/tests-list/test-details/test-details.component.spec.ts
gis727: Yes, awesome!

packages/frontend/src/app/repository/tests-list/test-details/test-details.component.ts
```diff
@@ -14,17 +14,37 @@
...
+      this.repoName,
+      this.test.name,
+      'https://flaky-dashboard.web.app/'
+    );
+    this.deleteUrl$.subscribe(res => {
+      console.info(res);
+      this.windowProvider.open(res, '_self');
+    });
+  }
 }
```
gis727, line 50: Since `deleteUrl$` is only used in `startDeleteTest`, it probably doesn't need to be an attribute of the component?
Also the observable can be immediately used without storing it at all. Like:
```
this.comService.fetchDeleteTestUrl(
 this.orgName,
 this.repoName,
 this.test.name,
 'https://flaky-dashboard.web.app/'
).subscribe(res => {
 this.windowProvider.open(res, '_self');
});
```

packages/frontend/src/app/repository/tests-list/test-details/test-details.component.ts
```diff
@@ -14,17 +14,37 @@
...
+
+  startDeleteTest() {
+    this.deleteUrl$ = this.comService.fetchDeleteTestUrl(
+      this.orgName,
+      this.repoName,
+      this.test.name,
+      'https://flaky-dashboard.web.app/'
+    );
+    this.deleteUrl$.subscribe(res => {
+      console.info(res);
```
gis727, line 46: We might want to remove the log? (users might be able to see it)

packages/frontend/src/app/services/com/com.service.spec.ts
```diff
@@ -216,6 +216,40 @@ describe('COMService', () => {
...
+    });
+
+    it('should call the error handler if an error occurs', () => {
+      const errorHandler = spyOn(service, 'handleError').and.returnValue(
+        empty()
+      );
+      const err = {} as HttpErrorResponse;
+      httpClientSpy.get.and.returnValue(throwError(err));
+
+      service.fetchTests('repo', 'org', []).subscribe();
```
gis727, line 248: I think `fetchDeleteTestUrl` is what we're testing here? (instead of `fetchTests`)

packages/api/server.js
gis727: Also, isn't 200 the default code in express? I think `res.send('...')` sends 200 as the status code by default?

packages/frontend/src/app/repository/tests-list/test-details/test-details.component.ts
```diff
@@ -14,17 +14,37 @@
...
   toPercentage(percentpassing: number): string {
     return (percentpassing * 100).toFixed(2);
   }
+
+  startDeleteTest() {
+    this.deleteUrl$ = this.comService.fetchDeleteTestUrl(
+      this.orgName,
+      this.repoName,
+      this.test.name,
+      'https://flaky-dashboard.web.app/'
```
gis727, line 43: Should this link come from `environment.baseUrl`?
Otherwise will it always redirect to the `production` website?

packages/frontend/src/app/services/com/com.service.spec.ts
cedpeters: Oops yes, thank you!

packages/frontend/src/app/repository/tests-list/test-details/test-details.component.ts
cedpeters: Oh perfect, thank you!

packages/api/server.js
```diff
@@ -23,21 +23,25 @@ const GetTestHandler = require('./src/get-test.js');
...
 const client = require('./src/firestore.js');
 const auth = require('./src/auth.js');
+const { InvalidParameterError } = require('./lib/errors');
 
 const { v4 } = require('uuid');
 
 const cors = require('cors');
 
-global.headCollection = process.env.HEAD_COLLECTION || 'testing/main/repos';
+global.headCollection = 'production/main/repos';
```
gis727, line 32: FYI, This will/must be reverted **before merging**.
Access to the production collection is necessary to test the Button in staging.

bcoe: 👍 let's not forget to revert this.

packages/api/server.js
```diff
@@ -23,21 +23,25 @@ const GetTestHandler = require('./src/get-test.js');
...
 app.use(bodyParser.json());
 app.use(bodyParser.urlencoded({ extended: true })); // support encoded bodies
 
-app.delete('/api/repo/:orgname/:reponame/test/:testid', async (req, res) => {
+app.get('/api/repo/deleteurl/:orgname/:reponame/test', async (req, res) => {
   const orgName = req.params.orgname;
   const repoId = req.params.reponame;
-  const testName = req.params.testid;
+  if (!req.query.testname) {
+    throw new InvalidParameterError('Route requires query parameter of name');
```
bcoe, line 42: I like that you're starting to build some helpers for different types of errors.

packages/api/server.js
```diff
@@ -81,13 +85,12 @@ app.get('/api/callback', async (req, res) => {
 
   const performed = await repo.performTicketIfAllowed(ticket, userPermission);
 
-  // todo send a message to the frontend to put a banner on the page indicating whether the action has been performed
   if (performed) {
     console.log('Successfully performed the action');
-    res.status(200).redirect(redirect);
+    res.status(200).redirect(redirect + ';done=' + performed);
```
bcoe, line 90: what values can performed take on?

packages/api/src/firestore.js~HEAD
```diff
@@ -0,0 +1,24 @@
+
+// Copyright 2020 Google LLC
```
bcoe, line 2: this file looks like it was potentially checked in by mistake?

packages/api/src/firestore.js~dd414e497455bb9dcc8aa1b7eea6ac76907e8a10
```diff
@@ -0,0 +1,24 @@
+
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
```
bcoe, line 8: likewise.

packages/api/src/repository.js
```diff
@@ -55,7 +55,7 @@ class Repository {
     return data.ticket;
   }
 
-  async allowedToPerformTicket (action, permission) {
```
bcoe, line 58: good catch 👍  all that `async` is, is a short hand for:

```
allowedToPerformTicket(() => {
  return new Promise((resolve, reject)) => {

  })
})
```

It's good to avoid specifying async unless it's needed.

packages/api/server.js
cedpeters: It's a boolean, returned by repo.performTicketIfAllowed. It will only be true if the test has been deleted.

PR #186, opened 2020-07-29
feat: add test deletion

cedpeters:
This PR actually calls the deleteTest method which deletes a test after authentication is finished.


packages/api/src/auth.js
```diff
@@ -36,17 +36,34 @@ class Auth {
...
+    if (userData.status !== 200) return null;
+
+    const data = await userData.json();
+    return data.login;
+  }
+
+  async retrieveUserPermission (accessToken, fullName) {
+    const login = await this.retrieveUserLogin(accessToken);
+
+    const result = await this.doFetch(`https://api.github.com/repos/${fullName}/collaborators/${login}/permission`, {
```
bcoe, line 54: awesome  👍 at the end of the day, I really like the route you've taken of calling APIs directly.

packages/api/test.tap
```diff
@@ -0,0 +1,109 @@
+1..76
+Example app listening at http://localhost:3000
+ok 1 Add-Build add-build Can add a build and repository to a blank collection
+BUIDLS>DATA: 1
+ok 2 Add-Build add-build Can add a different build with overlapping test cases and update repo info
+ok 3 Add-Build add-build Can add a third build with some new test cases, add same thing multiple times
```
bcoe, line 6: we shouldn't check in these tap files, not quite sure what happened here, as I added it to the `.gitignore`, I perhaps screwed up  the `.gitignore` directive.

Mind removing this with `git rm --force packages/api/test.tap`.

packages/api/test/auth-test.js
```diff
@@ -12,21 +12,31 @@
...
+    stubbed = sinon.stub(auth, 'doFetch').callsFake((link, object) => {
+      headers = object.headers;
+      fetchLink = link;
+      body = object.body;
+      return {};
+    });
+  });
+
+  after(() => {
+    stubbed.restore();
```
bcoe, line 36: love that you're cleaning up 😄

packages/api/test/repository-test.js
```diff
@@ -59,7 +64,99 @@ describe('Repository', () => {
...
+      stubbedTestDeletion = sinon.stub(deleter, 'deleteTest').callsFake((client, name, test) => {
+        fullName = name;
+        testName = test;
+      });
+    });
+
+    afterEach(() => {
+      stubbedPermissions.restore();
+      stubbedTestDeletion.restore();
+      fullName = null;
```
bcoe, line 145: this approach works, but I think you could simplify code simply by using:

```
spy.calledWith(arg1, arg2, ...);
```

basically, sinon has functionality built in similar to what you're doing here.

packages/api/src/auth.js
cedpeters: Thanks! I did look a bit into Github's octokit for interacting with their API, but for our needs it just seemed simplest to stick with direct API calls.

packages/api/src/repository.js
```diff
@@ -54,17 +55,33 @@ class Repository {
...
+  async performTicketIfAllowed (ticket, permission) {
+    const permitted = this.allowedToPerformTicket(ticket.action, permission);
+
+    if (permitted) {
+      if (ticket.action === 'delete-test') {
+        deleter.deleteTest(ticket.fullName, ticket.testName, client);
+      } else if (ticket.action === 'delete-repo') {
+        deleter.deleteRepo(client, ticket.fullName);
+      } else {
+        console.log('Invalid action');
```
bcoe, line 79: nit: might be worth making this `throw Error('invalid action')`, since this is really strange behavior and we'd like to bubble it.

packages/api/test/repository-test.js
```diff
@@ -100,22 +102,54 @@ describe('Repository', () => {
...
       assert.deepStrictEqual(result, true);
     });
   });
+
+  describe.only('performTicketIfAllowed', async () => {
+    let stubbedPermissions;
+    let stubbedTestDeletion;
+
+    beforeEach(() => {
+      stubbedPermissions = sinon.stub(repo, 'allowedToPerformTicket').callsFake((action, permission) => {
```
bcoe, line 129: I like this trick 👍

packages/api/test/repository-test.js
```diff
@@ -100,22 +102,54 @@ describe('Repository', () => {
...
+
+    afterEach(() => {
+      stubbedPermissions.restore();
+      stubbedTestDeletion.restore();
+    });
+
+    it('deletes test if permitted', async () => {
+      await repo.performTicketIfAllowed({ action: 'delete-test', fullName: 'org/repo', testName: 'test' }, 'permitted');
+
+      assert(stubbedTestDeletion.calledWith('org/repo', 'test'));
```
bcoe, line 145: lovely thanks for switching to `calledWith`.

PR #198, opened 2020-07-31
feat: added download button 

### LINKED ISSUE #195: Add download button feature in UX
- Add download feature that supports @StuartRucker backend work to download csv file

madelyndubuk:
-  Fixes #195 
- made settings icon only show up on repository page (thank you @gis727 for the help with the global service)
- added download button to be under settings icon
- removed api token/admin information since we are no longer using sessions
Screenshot 2020-07-31 at 11 17 24 AM


packages/frontend/src/app/services/globals/globals.service.ts
```diff
@@ -0,0 +1,41 @@
...
+  providedIn: 'root',
+})
+export class GlobalsService {
+  pageDataChange: BehaviorSubject<PageState> = new BehaviorSubject<PageState>({
+    currentPage: '',
+    repoName: '',
+    orgName: '',
+  });
+
+  update(pageName: string, orgName?: string, repoName?: string) {
```
bcoe, line 28: so the purpose of this global service, if I understand, is to maintain a global variable with a `repoName` and `orgName` variable.

What will `repoName` be populated with on the repo listing page, vs., the test listing page for a given repository.

packages/frontend/src/app/app-component/config/config.component.spec.ts
```diff
@@ -66,4 +70,24 @@ describe('ConfigComponent', () => {
...
+    component.envUrl = environment.baseUrl;
+    await fixture.detectChanges();
+
+    //click the export csv button
+    const downloadButton = fixture.debugElement.nativeElement.querySelector(
+      '.download-button'
+    );
+    downloadButton.click();
+    fixture.whenStable().then(() => {
+      expect(location.path()).toEqual(
```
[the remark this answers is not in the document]
madelyndubuk: I'm sorry that test is incorrect. I meant to convert to a draft, I am just trying to get this test to work then should be good

packages/frontend/src/app/app-component/app.component.ts
```diff
@@ -26,15 +26,21 @@ import {LicenseComponent} from '../license/license.component';
...
-        (event: NavigationEnd) => (this.showConfigWheel = event.url !== '/')
-      );
+    //get the current page's organization name and repository name
+    this.globals.pageDataChange.subscribe(pagedata => {
+      const onRepoPage =
+        pagedata.currentPage === RouteProvider.routes.repo.name;
+      this.showConfigWheel = onRepoPage;
+      this.repository = pagedata.repoName;
+      this.organization = pagedata.orgName;
+      console.log(onRepoPage);
```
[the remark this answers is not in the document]
madelyndubuk: no! I will get rid of it

packages/frontend/src/app/services/globals/globals.service.ts
madelyndubuk: Correct that is the purpose for the global service.  On the repository page it will have the name of the repository who's tests are currently being displayed.  That is a good question regarding the organization page that lists the repos because I am not 100% sure.  Good question for @gis727 ! I believe it should have an empty value for the repository at that point since the url for the org page does not have `:repo`, but I could be wrong.

packages/frontend/src/app/app-component/config/config.component.html
```diff
@@ -23,12 +23,17 @@
...
             <span *ngIf="showToken">hide</span>
         </a>
     </p>
 </div>
 
 <!-- NON-ADMIN VIEW -->
 <div id="non-admin-view" *ngIf="!showAdminView">
+     <!--DOWNLOAD BUTTON-->
+     <div id="download">
+        <a mat-raised-button class="download-button" href="{{this.envUrl}}/api/repo/{{this.orgName}}/{{this.repoName}}/csv" target="_blank">Export CSV</a>
```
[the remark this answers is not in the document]
gis727: Knowing that the user will be redirected anyway, yes, this is a nice way to handle this.

packages/frontend/src/app/app-component/config/config.component.spec.ts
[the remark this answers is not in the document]
gis727: 😅 According to a certain human, Magic is just unknown science. I don't think the harness is that powerful though.
To test the external redirection i think it can help to just extract the `window.location` somewhere like [here](https://github.com/GoogleCloudPlatform/flaky-service/pull/70/files#diff-43ee95303c542368da43e43342356dc4R39).

packages/frontend/src/app/services/globals/globals.service.ts
gis727: For the "repo listing page" AKA the org page, i think it'll be empty unless the user is searching a repository (in which case it will have that repository name).
On the test listing page AKA the repo page, it'll have the repository name.

packages/frontend/src/app/app-component/config/config.component.ts
```diff
@@ -12,30 +12,32 @@
...
-  showAdminView = false;
+  exportUrl = '';
+  windowProvider = window;
 
   constructor(public userService: UserService) {}
 
-  onTokenClick() {
-    this.showToken = !this.showToken;
+  onDownloadClick() {
+    this.exportUrl =
```
gis727, line 34: Since this is used only here, maybe we can create it here instead of having it as an attribute of the class ?
Like:
`const exportUrl = ...`

madelyndubuk: That is a good idea! I will do that instead, thank you

PR #215, opened 2020-08-04
Fix: 106% Bug

### LINKED ISSUE #211: ui: some tests pass 106% of the time
image

### LINKED ISSUE #214: Test labeled with "passes 0% of the time" after 1 success + 1 failure.
Probably related to #211.
## Observed behavior
A test that has been run 2 times with 1 failure is labeled with **Passes 0.00% of time**. Seems to be an issue with the metrics calculation.

Screenshot 2020-08-03 at 10 35 09 PM

## Expected behavior
The test should be labeled with **Passes 50.00% of the time**.

StuartRucker:
Fixed the bug described in this issue of some test cases having 106% passing. cacheSuccesses list includes all test cases, but mergedList only contains test cases in the last 15 which is where the discrepancy is from.

Bug was successfully duplicated and fixed. There were no test cases for inserting >15 builds due to time constraints which is why this slipped through.  

Fixes #211. Potentially fixes #214.


packages/api/lib/analytics.js
```diff
@@ -61,7 +61,13 @@ class TestCaseAnalytics {
   }
 
   computePassingPercent () {
-    return (this.mergedList.length === 0) ? 0 : this.cachedSuccess.length / this.mergedList.length;
+    let numSuc = 0;
+    this.mergedList.forEach(m => {
+      if (m.passed) {
+        numSuc += 1;
```
StuartRucker, line 67: I previously had this. Do you know why this wouldn't work?

```
    let numSuc = 0;
    for(let i = 0; i < this.mergedList.lenght; i ++){
      numSuc += this.mergedList[i].passed? 1: 0
    }
```
@bcoe

PR #223, opened 2020-08-06
feat: added new endpoint that requires action to do preprocessing

StuartRucker:
The new post endpoint `/api/build-gh` expects action to preprocess build metadata and send it in a flat structure. If any of the fields are missing, an error is safely thrown. Additionally, it makes sure that any fields that need to be escaped are escaped.

This is backwards compatible, but eventually the the original `api/build` endpoint will be removed. The `gh` stands for Github.


packages/api/src/post-build.js
```diff
@@ -192,6 +236,30 @@ class PostBuildHandler {
...
+        const testCases = PostBuildHandler.parseTestCases(parsedRaw, req.body.data);
+
+        const isValid = await validateGithub(req.body.metadata.token, decodeURIComponent(req.body.metadata.repoId));
+        if (!isValid) {
+          throw new UnauthorizedError('Must have valid Github Token to post build');
+        }
+
+        if (req.body.metadata.private) {
+          throw new UnauthorizedError('Flaky does not store tests for private repos');
+        }
```
tbpg, line 255: Can we move this up near the top to do as little work as possible if the repo is private?

packages/api/src/post-build.js
```diff
@@ -192,6 +236,30 @@ class PostBuildHandler {
         handleError(res, err);
       }
     });
+
+    this.app.post('/api/build-gh', async (req, res, next) => {
```
tbpg, line 240: Is there another name for this than `gh`? There might be test runners other than GitHub that send data this way.

tbpg, line 240: Will you add a comment above this explaining what this endpoint does, especially that part that is expected to already be done?

StuartRucker: renamed to v1. Then, every time a change is made the process can be adding new endpoint with higher version, updating action. And then after a year or so we can deprecate the old endpoint and remove that version to post to.

packages/api/src/post-build.js
```diff
@@ -167,6 +167,50 @@ class PostBuildHandler {
     return result;
   }
 
+  static cleanBuildInfo (metadata) {
+    const timestampNumb = Date.parse(metadata.timestamp);
+    const timestamp = isNaN(timestampNumb) ? new Date() : new Date(timestampNumb);
+
+    const returnVal = {
+      repoId: firebaseEncode(decodeURIComponent(metadata.repoId)),
+      organization: metadata.organization,
```
bcoe, line 176: what if I'm malicious and create an organization that's like `foo/johnny-drop-tables`. It seems like we might want to encode all fields for firebase, regardless of whether we think they could contain a `/`.

packages/api/src/post-build.js
```diff
@@ -192,6 +236,32 @@ class PostBuildHandler {
         handleError(res, err);
       }
     });
+
+    // endpoint expects the the required buildinfo to be in req.body.metadata to already exist and be properly formatted.
+    // required keys in the req.body.metadata are the inputs for addBuild in src/add-build.js
+    this.app.post('/api/build-v1', async (req, res, next) => {
```
bcoe, line 242: since this data is fairly GitHub specific, what if we used this opportunity to move to:

```
/api/build/gh/v1
```

Indicate that this is version one of our GitHub format upload?

packages/api/src/post-build.js
StuartRucker: I only escaped strings that will be used as keys in Firestore. the organization value will always be treated as a string. The only operation that it goes through are 
- a firestore update with organization: 'user input"
- a firestore read 
- a res.send of the json object read

PR #224, opened 2020-08-06
feat: endpoint to delete repo

cedpeters:


packages/api/server.js
```diff
@@ -59,6 +59,25 @@ app.get('/api/repo/deleteurl/:orgname/:reponame/test', async (req, res) => {
   res.status(200).send(url);
 });
 
+app.get('/api/repo/deleteurl/:orgname/:reponame', async (req, res) => {
```
bcoe, line 62: I might use `deleteurl` as the suffix, as it's the resource we're getting.

packages/api/lib/deleter.js
```diff
@@ -15,7 +15,7 @@
 
 const firebaseEncode = require('../lib/firebase-encode');
 
-async function deleteRepo (client, repoId) {
+async function deleteRepo (repoId, client) {
```
gis727, line 18: @bcoe, @cedpeters. The "flaky" test was actually deterministic :):
That line changed the signature of `deleteRepo`. But `deleteRepo` was used in `packages/api/test/get-batches.js`.
[Updating the signature](https://github.com/GoogleCloudPlatform/flaky-service/commit/e3f36970ea6e7646c2edea0d0d920538f8338b3a#diff-6b9db70b6a85df0db02734e6f54b81e0R365) fixed the test:

Issue #219, opened 2020-08-05
reduce size of  payload uploaded to GitHub


bcoe:
We are  getting an error [here](https://github.com/GoogleCloudPlatform/flaky-service/pull/212).

```bash
##[warning]Environment variable 'INPUT_GITHUB' exceeds the maximum supported length. Environment variable length: 36333 , Maximum supported length: 32766
```

I think we would be smart to potentially upload select fields. From a "reading the codebase"  point of view, this makes it easier to identify, through reading the action, which specific fields we care about from a test run.

This will help when we eventually support other services.

Issue #132, opened 2020-07-23
Consider moving the route for uploading GitHub action data to a different route


bcoe:
@StuartRucker and I talked about this today, as seen [here](https://github.com/GoogleCloudPlatform/flaky-service/pull/102) the data we accept on the post is becoming pretty specific to GitHub actions.

If we're going to stick with this approach, it would be good to perhaps have upload routes for specific build systems, e.g., `POST build/gh` rather than `POST build/`.

Issue #222, opened 2020-08-05
return error message to GitHub action if an attempt is made to upload privatedata


bcoe:
We should return a human readable error message that indicates why the upload was rejected, if a user attempts to upload test run information from a private repository.

Issue #93, opened 2020-07-18
action with error response exits with status 0


bcoe:
I would expect an action with an error response from the server to exit with `1` so that it fails the action:

<img width="886" alt="Screen Shot 2020-07-17 at 6 31 32 PM" src="https://user-images.githubusercontent.com/194609/87841654-cf141380-c85b-11ea-9a09-d87e63e74d24.png">

I think if folks then want to ignore the error, there's a way to specify this at the step level.

PR #226, opened 2020-08-07
fix: action Improvements - overflow error, messages, +

StuartRucker:
Moved the processing of upload metadata to the action from the server so the endpoint is more versatile and avoids overflow error. In conjunction with https://github.com/GoogleCloudPlatform/flaky-service/pull/223

- Fixes https://github.com/GoogleCloudPlatform/flaky-service/issues/132 by changing route to `/build/gh/v1` which creates a plan for future updates when there are users. A new endpoint will be created, without removing the old, that the newest version of the action can access
- Fixes https://github.com/GoogleCloudPlatform/flaky-service/issues/219 by not passing in the entire `github` context as environment variable
- https://github.com/GoogleCloudPlatform/flaky-service/issues/93 Addresses but does not necessarily "fix" this issue by adding more informative error messages that use the `core.warning` feature. Fails if action is used blatantly incorrectly. Warns if there is an issue such as an invalid token (so the rest of the steps in the action can continue)
- Uses environment variables when possible instead of passing in parameters

- Deletes an attempt at https://github.com/GoogleCloudPlatform/flaky-service/issues/222 (checking if repo is private will be done server side with Github token in the future)


packages/action/index.js
```diff
@@ -22,32 +22,78 @@ const core = require('@actions/core');
...
   } catch (error) {
-    console.log(error);
+    core.warning(error);
     core.setFailed(error.message);
   }
 }
 
 
-main();
+main();
```
tbpg, line 99: Newline at the end.

packages/action/index.js
```diff
@@ -22,32 +22,78 @@ const core = require('@actions/core');
 async function main() {
   try {
     // `who-to-greet` input defined in action metadata file
+    if(!core.getInput('repo-token') || !core.getInput('os') || !core.getInput('file-path') || !core.getInput('repo-token') ){
```
tbpg, line 25: Looks like `repo-token` is checked twice?

packages/action/index.js
```diff
@@ -22,32 +22,78 @@ const core = require('@actions/core');
...
+      throw "Github action missing required fields. Refer to documentation at https://github.com/googlecloudplatform/flaky-service";
```
tbpg, line 26: Can be a separate PR, but it doesn't look like the root README has relevant docs.

packages/action/index.js
```diff
@@ -22,32 +22,78 @@ const core = require('@actions/core');
...
+    }
+
+   
+    let buildmessageStr = process.env.GITHUB_WORKFLOW + ' ' + process.env.GITHUB_RUN_NUMBER;
```
bcoe, line 30: Nice, I'm glad we're building up this human readable string now.

packages/action/index.js
```diff
@@ -22,32 +22,78 @@ const core = require('@actions/core');
...
+      description: core.getInput('repo-description'),
+      buildmessage: buildmessageStr,
+      token: core.getInput('repo-token')
     };
-    const fileType = core.getInput('logtype');
-    console.log('reading from: ' + core.getInput('filepath'));
-    const data = fs.readFileSync(
-        core.getInput('filepath'), 'utf8');
 
+    metaData.matrix = JSON.stringify(metaData.environment.matrix, Object.keys(metaData.environment.matrix).sort()); //consistancy
```
bcoe, line 55: curious what the reasoning is for sending this to the API as a serialized string, vs., just leaving it an array. What's the ultimate representation in the database?

packages/action/index.js
```diff
@@ -22,32 +22,78 @@ const core = require('@actions/core');
...
       method: 'POST',
       body: JSON.stringify(sendMe),
       headers: {'Content-Type': 'application/json'},
     });
     const outcomeText = await outcome.text();
-    console.log('\nSERVER RESPONSE:');
-    console.log(outcomeText);
+    const outcomeAsJSON = JSON.parse(outcomeText);
+    if(outcomeAsJSON.error){
+      core.warning("Upload Failed - Status " + outcome.status);
```
bcoe, line 80: if we get an error, we should probably:

```
process.exitCode = 1
```

☝️ this will mean that the action indicates a failure.

packages/action/index.js
StuartRucker: In the database, its stored as a string. It keys need to be in alphabetic order because  in order to have a dropdown which filters by filter, you need to check for equality between the matrices. This could be done differently though. Another motivation for storing it as a string is to avoid a user input to be a Firestore key.

packages/action/index.js
StuartRucker: Okay sounds good. I initially avoided this because it could be annoying if your entire action gets cancelled because we are having server issues.

PR #241, opened 2020-08-12
refactor: optimizing reads/writes when writing test data

### LINKED ISSUE #221: figure out strategy for reducing writes
As things sit today, uploading a large test suite like Node.js (which consists of 2000 tests) is fairly expensive.

This is because each test  run results in 2000 writes. Could we reduce this cost by:

1. upfront, only performing writes  for tests that are currently failing or flaky?
2. perhaps, on a daily basis, calculating the overall metrics.

I'm open to other suggestions 😄 mainly, the  question is:

_given the majority of tests are,  hopefully, usually passing, can  we find a way to reduce writes for this category of test?_

CC: @StuartRucker

madelyndubuk:
- This PR attempts to improve the current system for uploading test information by minimize the number of writes to the Firestore collection storing tests
- Prior to these changes, our implementation for uploading data to Firestore involved two writes and one read per test run.  These operations would prove to be very costly if a large repository was consistently using Flaky.dev and uploading many builds a day (consider 50 builds a day for a repository with 2000 tests). 
- Our solution involves the following: we will now only update tests that are currently failing or have flaky behavior, therefore eliminating any writes to Firestore for tests that are currently passing and have been passing a lot recently (i.e. tests that seem to be consistently healthy)
- In every repository, all tests will be uploaded for the very first build
- The tests for a repository will now be stored in a Firestore collection called `queued`
-  This new implementation works with the assumption that builds will be uploaded chronologically (as currently is the case)
-  This new implementation also does not allow for an accurate passing percentage to be calculated for each test, since test deemed consistently healthy will not be tracked, therefore the new passing percentage is displayed along with the number of test runs that have been tracked and included in the calculations for the displayed passing percentage
Screenshot 2020-08-12 at 11 37 46 AM
Fixes #221


packages/api/lib/analytics.js
```diff
@@ -68,30 +69,81 @@ class TestCaseAnalytics {
...
 
-  // returns true if both...
-  // condition 1: there has been a failure within the last 5 runs
-  // condition 2: there have been any two failures separated by between 1 and 10 successes
-  computeIsFlaky () {
-    const failureIndices = [];
-    for (let k = 0; k < this.mergedList.length; k++) {
-      if (!this.mergedList[k].passed) {
-        failureIndices.push(k);
+  numberOfHistoricalTests () {
```
madelyndubuk, line 72: this is what will be used to indicate to users how many tests were involved in the calculation of the passing percentage that is surfaced to the UI

packages/api/lib/analytics.js
```diff
@@ -68,30 +69,81 @@ class TestCaseAnalytics {
...
+    return (this.mergedList.length);
+  }
+
+  // returns true if there have been any two failures separated by between 1 and 14 successes
+  calculateFlaky () {
```
madelyndubuk, line 77: a new function to compute a test's flakiness was needed, considering the fact that we are no longer keeping track of the past 15 test runs for every test in a repository

packages/api/lib/analytics.js
```diff
@@ -68,30 +69,81 @@ class TestCaseAnalytics {
...
       }
+    } else {
+      // test is not in the queue so NOT FLAKY
+      return false;
     }
-    if (failureIndices.length < 2) {
+  }
+
+  // function to see if we should start tracking subsequentpasses
+  calculateTrack () {
```
madelyndubuk, line 106: An additional function needed as part of the process of computing flakiness

packages/api/test/addbuild-getbuild.js
```diff
@@ -341,8 +351,7 @@ describe('Add-Build', () => {
 
       assert(ansObj[2].name === 'a/4'); // failed on build before the last build
 
-      assert(ansObj[3].name === 'a/3' || ansObj[3].name === 'a/1'); // both failed on most recent build
```
madelyndubuk, line 344: With the new implementation, not every test will be updated in the queue.  As in this case, test `a/1` was considered to be a consistently healthy test, so it will no longer show up

packages/api/test/analytics-test.js
```diff
@@ -94,44 +94,10 @@ describe('Flaky-Analytics', () => {
     it('Should track stats for tests', async () => {
       const testFlakyExpecations = [false, true, true, false];
       for (let i = 0; i < testFlakyExpecations.length; i++) {
-        const testData = await client.collection(global.headCollection).doc(buildInfo.repoId).collection('tests').doc(i.toString()).get();
+        const testData = await client.collection(global.headCollection).doc(buildInfo.repoId).collection('queued').doc(i.toString()).get();
         assert.strictEqual(testData.data().flaky, testFlakyExpecations[i]);
       }
     });
-
```
madelyndubuk, line 101: this test was removed considering we are assuming builds will be uploaded chronologically

packages/api/src/add-build.js
```diff
@@ -92,22 +101,67 @@ async function addTestRun (testCase, buildInfo, dbRepo) {
   );
 }
 
-async function updateTest (testCase, buildInfo, dbRepo) {
-  // first read the test
-  const prevTest = await dbRepo.collection('tests').doc(testCase.encodedName).get();
+async function updateQueue (isFirstBuild, testCase, buildInfo, dbRepo) {
```
madelyndubuk, line 104: Here is the new function that ultimately determines which tests we will be updating

madelyndubuk: @bcoe

madelyndubuk:
@googlebot I fixed it

packages/api/test/addbuild-getbuild.js
```diff
@@ -341,8 +348,8 @@ describe('Add-Build', () => {
...
-      assert(ansObj[4].name === 'a/3' || ansObj[4].name === 'a/1'); // both failed on most recent build
+      assert(ansObj[3].name === 'a/3'); // passed on build before the last build
+      assert(ansObj[4].name === 'a/1'); // has not been recently updated because considered healthy
```
madelyndubuk, line 352: the a/1 test has been considered healthy from the first build, so it will appear last in the list

madelyndubuk:
Fixes #221

packages/api/test.tap
```diff
@@ -0,0 +1,109 @@
+1..76
```
grayside, line 1: Should the tap file be included in the PR?

madelyndubuk: You are right, it shouldn't, thank you!

.github/workflows/ci.yaml
```diff
@@ -63,13 +64,14 @@ jobs:
       - name: track flaky tests
         continue-on-error: true
         if: ${{ always() }}
-        uses: GoogleCloudPlatform/flaky-service/packages/action@v1.0.0
```
StuartRucker, line 66: This looks like correct usage

Issue #144, opened 2020-07-24
Heatmap performance improvements


gis727:
The heatmap should fetch all builds for a specific day only when the user selects a day (not when the heatmap initializes).
This will require a new endpoint in the API and a refactoring of the current build-fetching endpoint.
So:
`api/repo/:orgname/:reponame/batches` would return all batches for a repo.
`/api/repo/:orgname/:reponame/batch/:timestamp` would return every build for the timestamp day.

Batches would be:

interface {
- **timestamp**: number,
- **failingBuilds**: number, `// # of failling builds for the day`
- **flakyBuilds**: number, `// ...`
- **passingBuilds**: number, `// ...`

}

bcoe:
@StuartRucker @gis727 I think what we'll probably want to do is pre-calculate the heat map, which just includes passed/failed counts on a given day, and an indication on a given day, if any of the tests at that point in time were considered flaky. Then, when someone clicks a given day, make a different API call that pulls in all the builds for that day.

PR #243, opened 2020-08-13
feat(API): builds on demand

gis727:
### Why this PR?
First step in the resolution of #144.

### What does this PR do?
This PR:
- Adds 2 endpoints to the API:
  - `/api/repo/:orgname/:reponame/batches` that returns all batches for a repository.
  - `/api/repo/:orgname/:reponame/batch/:timestamp` that returns all builds for the day corresponding to the `timestamp` parameter.
- Those 2 endpoints support the following parameters:
  - A `utcOffset` parameter that allows the API users to specify a custom offset that should be used when building the batches (defaults to 0).
  - `tag, ref, os, matrix` to filter the builds.

### Can this PR break anything in master?
No. Those endpoints are new (not used by the frontend) and are not dependant on any other PR.

### Can those changes be staged?
No. They aren't used by the frontend.


StuartRucker:
Nice improvement

PR #249, opened 2020-08-14
fix: Github Action handles endpoints ending in /

StuartRucker:
Used the `url` npm library `resolve` function to handle when the inputted URL ends in a slash, since you need to append the post endpoint `/api/build/gh/v1'


packages/action/package.json
```diff
@@ -13,7 +13,8 @@
     "@actions/core": "^1.2.4",
     "@actions/github": "^3.0.0",
     "node-fetch": "^2.6.0",
-    "tap-parser": "^10.0.1"
+    "tap-parser": "^10.0.1",
+    "url": "^0.11.0"
```
bcoe, line 17: actually  no  need  for  this dep, `url` is built into Node.js  with the exact same method signature.

StuartRucker: resolved

PR #254, opened 2020-08-18
feat: endpoint to post xunit test data

cedpeters:


packages/api/lib/xunit-parser.js
```diff
@@ -0,0 +1,88 @@
...
+      // Ignore skipped tests. They didn't pass and they didn't fail.
+      if (testcase.skipped !== undefined) {
+        continue;
+      }
+
+      const failure = testcase.failure;
+      const error = testcase.error;
+
+      const testCaseRun = new TestCaseRun((failure === undefined && error === undefined) ? 'ok' : 'not ok', count, (testsuiteName !== '') ? testsuiteName + '/' + testcase._attributes.name : testcase._attributes.name);
+      count++;
```
bcoe, line 67: I'm a little concerned that `count` might not be consistent between test runs, depending on the test runner, i.e., the same test might not fall in the same place in the output. This is carried over from tap tests, so nothing to change right now.

packages/api/lib/xunit-parser.js
```diff
@@ -0,0 +1,88 @@
...
+    // If there were no tests in the package, continue.
+    if (testcases === undefined) {
+      continue;
+    }
+    // If there is only one test case, put it into an array to make it iterable.
+    if (!Array.isArray(testcases)) {
+      testcases = [testcases];
+    }
+
+    for (const testcase of testcases) {
```
bcoe, line 57: Rather than assigning `count` outside of the loop, since it's only used in the loop, I would be tempted to just define the counter here:

```
for (let testcase, count = 1; (testcase = testcases[count - 1]) !== undefined; count++) {
  // ...
}}
```

packages/api/src/post-build.js
```diff
@@ -121,7 +122,7 @@ class PostBuildHandler {
     switch (fileType) {
       case 'TAP': {
         var data = [];
-        var p = new Parser();
+        var p = new TapParser();
```
bcoe, line 125: I like this clarification of what type of parser is being used here.

packages/api/src/post-build.js
```diff
@@ -18,7 +18,8 @@
 
 const addBuild = require('../src/add-build');
 const TestCaseRun = require('../lib/testrun');
-var Parser = require('tap-parser');
+var TapParser = require('tap-parser');
```
bcoe, line 21: use `const` rather than `var` 👍 (since we won't be reassigning this variable).

packages/api/lib/xunit-parser.js
cedpeters: I actually wound up completely deleting the count var from testrun.js.

packages/api/lib/xunit-parser.js
cedpeters: Yeah, turns out it is never actually a useful value and it isn't stored in the database. I removed it entirely.

packages/api/lib/xunit-parser.js
```diff
@@ -0,0 +1,128 @@
...
+    }
+    return tests;
+  }
+
+  // IMPORTANT: All values that will be used as keys in Firestore must be escaped with the firestoreEncode function
+  cleanXunitBuildInfo (metadata) {
+    return {};
+    //    const timestampNumb = Date.parse(metadata.timestamp);
+    //    const timestamp = isNaN(timestampNumb) ? new Date() : new Date(timestampNumb);
+
```
bcoe, line 90: looks like this commented out code can be removed?

packages/api/lib/xunit-parser.js
```diff
@@ -0,0 +1,128 @@
...
+    }
+    // If there is only one test suite, put it into an array to make it iterable.
+    if (!Array.isArray(testsuites)) {
+      testsuites = [testsuites];
+    }
+    for (const suite of testsuites) {
+    // Ruby doesn't always have _attributes.
+      let testsuiteName = suite._attributes ? suite._attributes.name : undefined;
+
+      // Get rid of github.com/orgName/repoName/
```
azizsonawalla, line 38: You could probably use regex here for this: 

```javascript
function trim(url) {
  return url.replace(/(.)*github.com\/[a-zA-Z\-]*\/[a-zA-Z\-]\/*/g, '');
}

console.log(trim("https://github.com/GoogleCloudPlatform/flaky-service/testName")); // prints "testName"
console.log(trim("github.com/GoogleCloudPlatform/flaky-service/testName")); // prints "testName"
```

packages/api/src/post-build.js
```diff
@@ -31,6 +32,10 @@ class PostBuildHandler {
     this.client = client;
   }
 
+  static async addBuild (testCases, buildInfo, client, collectionName) {
+    await addBuild(testCases, buildInfo, client, collectionName);
+  }
```
azizsonawalla, line 37: Is there a reason for this static method if it's just calling another function?

packages/api/src/post-build.js
```diff
@@ -96,7 +101,7 @@ class PostBuildHandler {
       if (typeof x.ok !== 'boolean' || !x.id || !x.name) {
         throw new InvalidParameterError('Missing All Test Case Info');
       }
-      const testcase = new TestCaseRun(x.ok ? 'ok' : 'not ok', x.id, x.name);
+      const testcase = new TestCaseRun(x.ok ? 'ok' : 'not ok', x.name);
```
azizsonawalla, line 104: You seem to be using `"ok"` and `"not ok"` strings in multiple places -- might be helpful to save them as constants and reuse the constant

packages/api/lib/xunit-parser.js
cedpeters: Ah, thank you! Just fixed that.

packages/api/src/post-build.js
cedpeters: Yes, it's much simpler to stub addBuild this way. Alternatively, I believe, we could have addBuild `export {addBuild}`. This would allow us to stub the original function, though then every time we used it we have to do something like `addBuildModule.addBuild`. @bcoe would you prefer to keep it this way or should I change it?

Here's an article I found about exactly this situation: https://medium.com/@minaluke/how-to-stub-spy-a-default-exported-function-a2dc1b580a6b

packages/api/lib/xunit-parser.js
```diff
@@ -0,0 +1,92 @@
...
+      for (const testcase of testcases) {
+      // Ignore skipped tests. They didn't pass and they didn't fail.
+        if (testcase.skipped !== undefined) {
+          continue;
+        }
+
+        const failure = testcase.failure;
+        const error = testcase.error;
+
+        const testCaseRun = new TestCaseRun((failure === undefined && error === undefined) ? 'ok' : 'not ok', (testsuiteName !== '') ? testsuiteName + '/' + testcase._attributes.name : testcase._attributes.name);
```
JustinBeckwith, line 60: This line of code is difficult to read.  Could we break this down into a few lines?  (they're free!)

packages/api/lib/xunit-parser.js
```diff
@@ -0,0 +1,92 @@
...
+          testCaseRun.failureMessage = log;
+        }
+
+        tests.push(testCaseRun);
+      }
+    }
+    return tests;
+  }
+
+  trim (url) {
```
JustinBeckwith, line 79: For each method here - can we get a jsdoc header that explains what's happening?

packages/api/src/post-build.js
cedpeters: I think this is actually a quirk of parsing tap files, I don't think I can change it.

packages/api/test/post-build-test.js
```diff
@@ -261,6 +262,58 @@ describe('Posting Builds', () => {
     assert.strictEqual(result.data().environment.ref, 'master');
   });
 
+  describe('xunit endpoint', async () => {
+    let stubs = [];
+
+    afterEach(() => {
+      /** Cleanup **/
+      stubs.forEach(stubbed => {
```
JustinBeckwith, line 270: the pattern I usually use here is to create a sandbox, and then in the `afterEach` call `sandbox.restore()`.  That would help you skip keeping track of all your used stubs:

```js
const sandbox = sinon.createSandbox();
..
const stub1 = sandbox.stub(obj, 'method')....
const stub2 = sandbox.stub(obj, 'method')....

# in your afterEach
sandbox.restore();
```

packages/api/src/post-build.js
bcoe: I agree with @azizsonawalla that the approach we took of encapsulating the method in a `static` method reads a little odd ... I don't have a strong opinion though, I like how easy it makes our mocking.

If @azizsonawalla makes a case for it, I think a reasonable alternative would be that we switch to two exports like the article you shared suggests, so:

```
module.exports = {
  addBuild,
  PostBuildHandler
}
```

I don't think we should pull `proxyquire` into the project, as we're already using `nock` and `sinon`.

packages/api/test/post-build-test.js
cedpeters: Oh yeah that's much more elegant, thank you!

packages/api/src/post-build.js
cedpeters: I just went ahead and tried having `add-build` export `{addBuild}` instead of `addBuild`, and it took a bit of refactoring but it works now without that little function.

PR #263, opened 2020-08-23
passing boolean instead of ok/not ok string for testrun constructor

### LINKED ISSUE #258: Get rid of `ok, not ok` in testrun.js constructor
The `testrun` constructor (https://github.com/GoogleCloudPlatform/flaky-service/blob/master/packages/api/lib/testrun.js) accepts a string, which is either `ok` or `not ok`, and then immediately converts it into a boolean. Fix it so that it just accepts a boolean in the first place. This will involve changing all the places where the constructor is called.

rbrishabh:
Fixes #258 

Hey @cedpeters,  please let me know if this needs any changes and I'll be happy to make them!

Thanks!


bcoe:
@rbrishabh thank you for the contribution 😄 

@cedpeters let me know when you're ready with review? I believe tests are failing due to a permission issue with GitHub actions, nothing to do with the PR.

cedpeters:
@bcoe sounds good! As far as I'm concerned it looks good to go, my nits are just style suggestions. I can make a small PR for fixing the invalid constructor call in tap-parser.js, because that was my mistake and not introduced in this PR.

rbrishabh:
If you want, I can quickly push both those changes in this PR only @cedpeters

rbrishabh:
Actually, I'm on it right now! Happy to contribute :)

cedpeters:
I wrote a small PR to fix the bug I introduced, but it's having some issues with the cla that I'll need to figure out. Might just be simpler for you to do it. https://github.com/GoogleCloudPlatform/flaky-service/pull/264

rbrishabh:
@cedpeters this is ready for another review! Thanks!

packages/api/src/post-build.js
```diff
@@ -97,7 +97,7 @@ class PostBuildHandler {
       if (typeof x.ok !== 'boolean' || !x.id || !x.name) {
         throw new InvalidParameterError('Missing All Test Case Info');
       }
-      const testcase = new TestCaseRun(x.ok ? 'ok' : 'not ok', x.name);
+      const testcase = new TestCaseRun(x.ok ? true : false, x.name);
```
cedpeters, line 100: Might want to do it like `const testcase = new TestCaseRun(x.ok, x.name);`

packages/api/test/analytics-test.js
```diff
@@ -64,7 +64,7 @@ describe('Flaky-Analytics', () => {
     for (let i = 0; i < testCases.length; i++) {
       const testCaseObjs = [];
       for (let k = 0; k < testCases[i].length; k++) {
-        testCaseObjs.push(new TestCaseRun(testCases[i][k] ? 'ok' : 'not ok', k, k.toString()));
+        testCaseObjs.push(new TestCaseRun(testCases[i][k] ? true : false, k, k.toString()));
```
cedpeters, line 67: Too many args in this one, too, sorry. Should be: `testCaseObjs.push(new TestCaseRun(testCases[i][k], k.toString()));`

bcoe:
@rbrishabh sounds like this is ready to land, as soon  as you address @cedpeters' last comment and resolve the merge conflict 👍

rbrishabh:
@bcoe I think this is ready

bcoe:
@rbrishabh thank you for the contribution 👌

rbrishabh:
My pleasure! I'll be happy to keep contributing @bcoe, so if you think there's anything I can do to help, just tag @rbrishabh and I'll be there!

PR #264, opened 2020-08-23
fix: remove bad constructor calls

cedpeters:
Just a tiny fix from a refactor I did last week.

Might have a small merge conflict with https://github.com/GoogleCloudPlatform/flaky-service/pull/263. Once that one is landed, this PR may be unnecessary but if it isn't then I will fix the conflict.


cedpeters:
@googlebot I signed it!

PR #265, opened 2020-08-24
fix(api): addressed bugs with queue logic for tests

bcoe:
This PR fixes a few issues with the test tracking queue logic:

##  What this fixes

1. subsequent passes was never being incremented.
2. searchindex was always being calculated as `2`, it should be: `2 = falure`,  `1 = flaky`,  `0 = passing`.
3. the flaky logic and "subsequent pass" logic was being conflated, the logic now is that: flakiness is calculated independently from pass count; if we see N passes in a row, we stop updating the test.

## Some small improvements made in PR

1. we  were not deleting stale data from firestore, we now  do with `cleanupStaleTestData`.
2. I started a `test/unit` folder, which contains tests that do not require a `Firestore` connection, ideally we'll move more testing to this approach over time.


packages/api/test/analytics-test.js
```diff
@@ -59,7 +59,7 @@ const testCases = [
 
 describe('Flaky-Analytics', () => {
   before(async () => {
-    global.headCollection = 'testing/' + 'analytics-testsuite-' + uuidv4() + '/repos'; // random collection name for concurrent testing
+    global.headCollection = 'testing/' + Date.now() + '-analytics-testsuite-' + uuidv4() + '/repos'; // random collection name for concurrent testing
```
bcoe, line 62: I added a `timestamp` as a prefix to our collections, so that we can use it to decide if data is stale when deleting.

packages/api/test/repository-test.js
```diff
@@ -12,161 +12,40 @@
...
+      await repo.deleteDoc(`testing/${uniqueString}`);
     });
 
     it('returns null when a document does not exist', async () => {
       const data = await repo.getDoc('doesnt-exist/super-fake');
       assert.deepStrictEqual(data, null);
     });
   });
-
-  describe('getCollection', async () => {
```
bcoe, line 54: moved these tests into `test/unit`, since they don't need a database.

bcoe:
@cedpeters I"m going to go ahead and land this, as it fixes a few things, but please feel free to follow up with some review.

packages/api/src/repository.js
```diff
@@ -30,27 +30,18 @@ class Repository {
...
-    const snapshot = await client.collection(`${identifier}`).get();
-    if (snapshot.empty) {
-      return result;
-    }
-    snapshot.forEach(doc => {
-      const entry = doc.data();
-      result.push(entry);
-    });
-    return result;
+  ticketCollection () {
```
cedpeters, line 33: This is really good! I agree it's better not to rely on hard-coding by default.

packages/api/test/analytics-test.js
cedpeters: Makes sense! I can't remember, but we track the timestamp for individual tests' runs as well, right? I'd thought about checking the runtime on one of the tests to decide if data is stale, but this seems much more straightforward.

packages/frontend/src/app/heat-map/heat-map.component.spec.ts
```diff
@@ -217,6 +217,10 @@ describe('HeatMapComponent', () => {
   }));
 
   it('should show the tooltip when the mouse hovers a cell', fakeAsync(() => {
+    mockCOMService.fetchBatches = () => of(mockBatches.none);
```
cedpeters, line 220: Oh good, looks like you fixed those flaky tests! That's such a relief, they were getting pretty annoying.

PR #268, opened 2020-08-24
feat: support xunit from pubsub subscription

### LINKED ISSUE #262: Middleware to block xunit parser when in production
We discussed deploying a second copy of Flaky.dev which could be used for Google-managed repositories. If we do this, we should add middleware that checks the environment and blocks attempts to post to the xunit endpoint (https://github.com/GoogleCloudPlatform/flaky-service/issues/261) unless it's the corporate version.

bcoe:
This sees @cedpeters' work over the finish line, making it so we can accept data posted directly from the PubSub subscription used by our build cop bot.

The data posted from [PubSub](https://cloud.google.com/pubsub) just happens to be a slightly different shape than we planned for:

```json
{
  "message": {
     "data": "base64data"
  }
}
```

Otherwise I believe we can use the work as implemented.

fixes #262, #261


packages/api/lib/parse-pubsub-payload.js
```diff
@@ -0,0 +1,48 @@
...
+    buildId: match ? match.groups.url : metadata.commit,
+    sha: metadata.commit,
+    name: repo,
+    description: '',
+    buildmessage: ''
+  };
+}
+
+module.exports = (pubsubPayload) => {
+  assert(pubsubPayload.message, 'pubsub payload did not contain message field');
```
cedpeters, line 40: I've never seen assert used outside of a test class like this. I'm curious, what does this do? It seems to *require* that the message fields be left empty?

packages/api/lib/xunit-parser.js
```diff
@@ -93,21 +89,7 @@ class Parser {
...
-  /**
-  * Parse the build data and convert it into a JSON format that can easily be read and stored.
-  * @param metadata Contains all data sent to the endpoint besides the actual test results.
-  * @returns {JSON} The metadata for the build
-  */
-  cleanXunitBuildInfo (metadata) {
-    metadata.repoId = firebaseEncode(metadata.repoId);
-    metadata.buildId = firebaseEncode(metadata.buildId);
-    return metadata;
+    return url.replace(/^github.com\/[^/]+\/[^/]+\/?/, '');
```
cedpeters, line 92: This is so much better than my mangled attempt at regex!

packages/api/lib/parse-pubsub-payload.js
```diff
@@ -0,0 +1,48 @@
...
+  const [org, repo] = metadata.repo.split('/');
+  const match = metadata.buildURL.match(/\[Sponge\]\((?<url>.*)\)/);
+  return {
+    repoId: firebaseEncode(decodeURIComponent(metadata.repo)),
+    organization: org,
+    timestamp: new Date(),
+    url: match ? match.groups.url : `https://github.com/${metadata.repo}`,
+    environment: {
+      matrix: '{"kokoro":""}',
+      os: 'Linux',
```
cedpeters, line 27: Should we make a note somewhere that this is essentially fake data? Or instead of hard-coding 'Linux' we could say `os: 'undefined'` or something like that?

packages/api/lib/parse-pubsub-payload.js
bcoe: @cedpeters as in tests, it will throw an exception if the statement doesn't evaluate to `true`.

It can be a good way to make an error related to inputs more clear, otherwise the exception might occur later in the program because a field can't be found on an object you were  expecting.

packages/api/lib/xunit-parser.js
bcoe: @cedpeters practice makes perfect 😄 regex is a great thing to keep getting better at.

packages/api/lib/parse-pubsub-payload.js
bcoe: I think a comment would be good, I believe our servers are all `Linux`, but that's obviously not the case  for someone else using this.

bcoe:
@cedpeters a few bugs aside, I was able to wire up your XML parsing and it's  working, FYI:

Screen Shot 2020-08-24 at 10 06 46 AM

Screen Shot 2020-08-24 at 10 06 40 AM

packages/api/lib/parse-pubsub-payload.js
cedpeters: @bcoe Ah okay, thank you! I'm curious why we specifically demand that the pubsub must not contain a message field. Would we run into problems if we just asserted that the message has to exist, and not that it necessarily must be that specific one?

cedpeters:
@bcoe That's amazing! So cool to see it in action at last!

